PR #1, opened 2020-10-13
New parameter out_all, default false

traylenator:
In order to allow all outbound traffic a parameter is
added to enable a simple `allow` entry on the out chain.

Default is false so backwards compatible.


traylenator:
I was unsure how you like to do parameter testing in spec files , happy to follow another strategy if you  can point me at one.

manifests/ip_nat.pp
```diff
@@ -51,4 +51,7 @@
   if $nftables::out_https {
     include nftables::rules::out::https
   }
+  if $nftables::out_all {
```
duritong, line 54: We could simplify the generated chain, by leaving away all the other out if that one is set.

manifests/rules/out/all.pp
```diff
@@ -0,0 +1,7 @@
+# allow all outbound
+class nftables::rules::out::all {
+  nftables::rule{
+    'default_out-all':
+      content => 'accept',
```
duritong, line 5: this should probably have a high order like 90, so one can still easily put deny statements in front.

manifests/ip_nat.pp
traylenator: For sure that could make sense and I configured as :+1: 

```yaml
nftables::out_ntp:   false
nftables::out_dns:   false
nftables::out_http:  false
nftables::out_https: false
nftables::out_all:   true
```
would you consider accepting to change the default to false for all of the above? This would make the code cleaner and more obvious at the expense of backwards compatibility.

manifests/rules/out/all.pp
traylenator: Yes makes sense, will change when I take action on first one above.

manifests/ip_nat.pp
duritong: So we added these by default, as this is what we need to open +- on every system. So it's more convenient to get it running, by just including the nftables class. The result is that you have a locked down system, that is still able to talk to the outside for the things it need (time, dns + yum repos) and all the other rules you might be added by classes or we should define a more generic way to specify rules via hiera at some point.

Now I am not sure how the code would look easier if these are false by default? The hieradata would, but it also would look easier if we only include the out rules, if out_all is false.

duritong: Meaning:

```puppet
# basic outgoing rules
if $nftables::out_all {
  include nftables::rules::out::all
} else {
  if $nftables::out_ntp {
    include nftables::rules::out::chrony
  }
  if $nftables::out_dns {
    include nftables::rules::out::dns
  }
  if $nftables::out_http {
    include nftables::rules::out::http
  }
  if $nftables::out_https {
    include nftables::rules::out::https
  }
}
```

traylenator: Sure thing , I'll write something that disables the out rules if `out_all` is enabled.

traylenator: Exactly as you suggested - seems good.

traylenator:
I started adding some strings docs as well.

duritong:
Could you rebase on latest master?

@keachi PTAL

manifests/init.pp
```diff
@@ -1,10 +1,38 @@
...
+#   Allow outbound to http servers.
+#
+# @param out_https
+#   Allow outbound to https servers.
+#
+# @param out_https
+#   Allow outbound to https servers.
+#
+# @param in_ssh
+#   Allow outbound to https servers.
```
keachi, line 27: This should be inbound

keachi (approved):
LGTM

traylenator:
rebased and squashed also.

duritong:
Awesome, thank you!

Issue #2, opened 2020-11-10
The fate of forbidden packets should be configurable


LionelCons:
A forbidden packet is currently logged and denied. This is hard-coded in the 'INPUT-log_rejected' fragment and it is not configurable.

The fate of forbidden packets should be configurable.

This is especially important for the log prefix and the associated action (some may want to drop rather than reject).

duritong:
Sounds reasonable, happy to merge a MR

LionelCons:
FWIW, I have tested Nacho's `log_prefix` and `reject_with` and they work fine in my environment. So these 2 MRs do fix this issue.

PR #5, opened 2020-11-15
Allow customising the log prefix

nbarrientos:
This patch partly implements #2, allowing the caller of the module to configure the prefix that's used to log when packets reach the end of the chains. Perhaps the default values of the parameters of the `nftables` class should be moved to module-level Hiera data (`data/common.yaml`) to make it cleaner.


duritong:
Thank you. Not sure about the data extraction, personally I like defaults in the code and only e.g. moving per distro defaults out into data. But well, that's not really a strong opinion.

PR #6, opened 2020-11-16
Add rules for afs3_callback in and out rules for kerberos and openafs.

traylenator:
Add the afs callback to the cache manager(7001) which is UDP and always
IPv4 since there OpenAFS does not support IPv6.

```puppet
include nftables::rules::afs3_callback
```

For background OpenAFS servers connect to clients on the `afs3_callback` port and it is the server
initiating the connection. The clients have already connected to server on a completely different ports.

To configure an openafs client in totality 

```puppet
include nftables::services::openafs_client
```
which will allow both outbound connections clients need and the inbound cache callbacks. 

https://wiki.openafs.org/devel/AFSServicePorts/

### DIFF
--- a/manifests/rules/afs3_callback.pp
+++ b/manifests/rules/afs3_callback.pp
@@ -1,0 +1,16 @@
+# @summary Open call back port for AFS clients
+# @param saddr list of source network ranges to a
+# @example
+# class{'nftables::rules::afs3_callback':
+#   saddr => ['192.168.0.0/16', '10.0.0.222']
+# }
+#
+class nftables::rules::afs3_callback (
+  Array[Stdlib::IP::Address::V4,1] $saddr = ['0.0.0.0/0'],
+) {
+
+  nftables::rule{'default_in-afs3_callback':
+    content =>  "ip saddr { ${saddr.join(', ')} } udp dport 7001 accept";
+  }
+
+}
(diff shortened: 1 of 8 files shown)


traylenator:
Happy to add an `nftables::rules::out::afs_client` though to be honest it is not something we have real world experience of.

traylenator:
@nbarrientos and I are happy to do some reviews of other MRs by the way.

duritong:
> Happy to add an `nftables::rules::out::afs_client` though to be honest it is not something we have real world experience of.

If you have at least an idea, what should be added, it would be nice to have it.

traylenator:
@duritong I checked with one of the AFS admins who advised those ports. It may well differ for different  cells but that's okay.

traylenator:
While writing the tests for `nftables::rules` I realised rule title can have a suffix on the end.
What was the idea here?

Should I use `-1` and `-2` for the kerberos udp and tcp rules.

On a related note I see there is a udp+tcp in one line but it is pretty unreadable.
```
inet filter input meta l4proto {tcp, udp} @th,16,16 { 88 }
```
can switch of course.

duritong (approved):
thank you!

keachi (approved):
lgtm

PR #7, opened 2020-11-16
Add a parameter to control the fate of discarded traffic

### LINKED ISSUE #2: The fate of forbidden packets should be configurable

nbarrientos:
This patch implements the second feature requested via #2, adding a parameter to configure how discarded packets will be handled. By default, they'll be rejected by the last rule in each chain with `icmpx type port-unreachable` as it's always been, however it's possible now to set the parameter to `false` so the packets are handled instead via the policies of the chains which at the moment are all `drop`.

Closes #2.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -26,15 +26,24 @@
 # @param in_ssh
 #   Allow inbound to ssh servers.
 #
+# @param reject_with
+#   How to discard packets not matching any rule. If `false`, the
+#   fate of the packet will be defined by the chain policy (normally
+#   drop), otherwise the packet will be rejected with the REJECT_WITH
+#   policy indicated by the value of this parameter.
+#
 class nftables (
-  Boolean $in_ssh    = true,
-  Boolean $out_ntp   = true,
-  Boolean $out_dns   = true,
-  Boolean $out_http  = true,
-  Boolean $out_https = true,
-  Boolean $out_all   = false,
-  Hash $rules        = {},
-  String $log_prefix = '[nftables] %<chain>s Rejected: ',
+  Boolean $in_ssh                = true,
+  Boolean $out_ntp               = true,
+  Boolean $out_dns               = true,
+  Boolean $out_http              = true,
+  Boolean $out_https             = true,
+  Boolean $out_all               = false,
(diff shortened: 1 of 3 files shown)


duritong:
LGTM @keachi wdyt

duritong (approved):

manifests/init.pp
```diff
@@ -26,15 +26,24 @@
...
-  Hash $rules        = {},
-  String $log_prefix = '[nftables] %<chain>s Rejected: ',
+  Boolean $in_ssh                = true,
+  Boolean $out_ntp               = true,
+  Boolean $out_dns               = true,
+  Boolean $out_http              = true,
+  Boolean $out_https             = true,
+  Boolean $out_all               = false,
+  Hash $rules                    = {},
+  String $log_prefix             = '[nftables] %<chain>s Rejected: ',
```
keachi, line 43: I suppose to move `Rejected` to the `inet_filter.pp`. If the log prefix is used at another place where the packages are just logged and not rejected, the prefix is wrong. And the variable name is really generic.

keachi (changes requested):
Just one little nit picking.

manifests/init.pp
nbarrientos: Thanks for your comment. Not sure I'm following, though. Are you suggesting to replace the default value of `$log_prefix` here by `[nftables] %<chain>s` and perhaps append the `Rejected:` bit [here](https://github.com/duritong/puppet-nftables/blob/master/manifests/inet_filter.pp#L61) where we now for sure that the packet will be discarded (either via a reject or a drop)? Makes sense.

nbarrientos: If so, would you mind if I did these changes in a separate MR?

keachi: Yes, that's what I meant. Thank you for asking and clarification. You can implement that in another merge request, that's fine for me.

PR #8, opened 2020-11-16
Allow raw sets and dashes in set names

nbarrientos:
This patchset does two things:

* Relax the parameter `type` of the defined type `nftables::set` so it does not have to be always set, allowing to define a set passing raw input using `source` or `content` (see tests)
* Allow dashes to be part of set names.

### DIFF
--- a/manifests/set.pp
+++ b/manifests/set.pp
@@ -2,7 +2,7 @@
 define nftables::set(
   Enum['present','absent']
     $ensure = 'present',
-  Pattern[/^[a-zA-Z0-9_]+$/]
+  Pattern[/^[-a-zA-Z0-9_]+$/]
     $setname = $title,
   Pattern[/^\d\d$/]
     $order = '10',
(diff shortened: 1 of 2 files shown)


manifests/set.pp
```diff
@@ -1,13 +1,13 @@
 # manage a named set
 define nftables::set(
-  Enum['ipv4_addr', 'ipv6_addr', 'ether_addr', 'inet_proto', 'inet_service', 'mark']
-    $type,
   Enum['present','absent']
     $ensure = 'present',
-  Pattern[/^[a-zA-Z0-9_]+$/]
+  Pattern[/^[-a-zA-Z0-9_]+$/]
```
duritong, line 5: shouldn't that be escaped `\-` ?

duritong (changes requested):
Other than the minor comment about escaping, all good.

manifests/set.pp
nbarrientos: Hi! Typically, inside character classes the hyphen can be added at the beginning or at the end unescaped as the context is enough for the parser to remove the special meaning.

duritong: I see, didn't know about that.

nbarrientos:
Thanks for the review!

PR #10, opened 2020-11-18
Reload rules atomically and verify rules before deploy

traylenator:
Background: The unit file for nftables on CentOS 8 contains:

```
ExecStart=/sbin/nft -f /etc/sysconfig/nftables.conf
ExecReload=/sbin/nft 'flush ruleset; include "/etc/sysconfig/nftables.conf";'
ExecStop=/sbin/nft flush ruleset
```

As things stood on config modification `systemctl stop nftables ; systemctl start nftables` was being
called resulting in:

```
nft flush ruleset
nft -f /etc/sysconfig/nftables.conf
```

as distinct commands so a non-atomic flush and load of ruleset.

With this change it is now.

```
/sbin/nft 'flush ruleset; include "/etc/sysconfig/nftables.conf";'
```
Also added is validation of the ruleset by puppet prior to updating the real configuration.

Configuration is deployed to `/etc/nftables/puppet-preflight/` and `/etc/nftables/puppet-preflight.nft`

This is validate with `nft -c` and if and only if the configuration is valid in the nft sense the configuration
will be copied to the live location `/etc/nftables/puppet/` and `/etc/nftables/puppet.nft` before the service
is reloaded.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -84,6 +84,24 @@ class nftables (
       owner => 'root',
       group => 'root',
       mode  => '0640';
+    '/etc/nftables/puppet-preflight':
+      ensure  => directory,
+      mode    => '0750',
+      purge   => true,
+      force   => true,
+      recurse => true;
+    '/etc/nftables/puppet-preflight.nft':
+      ensure => file,
+      source => 'puppet:///modules/nftables/config/puppet.nft';
+  } ~> exec{
+    'nft validate':
+      refreshonly => true,
+      command     => '/usr/sbin/nft -I /etc/nftables/puppet-preflight -c -f /etc/nftables/puppet-preflight.nft || ( /usr/bin/echo "#CONFIG BROKEN" >> /etc/nftables/puppet-preflight.nft && /bin/false)';
+  } -> file{
+    default:
+      owner => 'root',
+      group => 'root',
+      mode  => '0640';
     '/etc/nftables/puppet.nft':
       ensure => file,
       source => 'puppet:///modules/nftables/config/puppet.nft';
@@ -94,8 +112,17 @@ class nftables (
       force   => true,
       recurse => true;
   } ~> service{'nftables':
-    ensure => running,
-    enable => true,
+    ensure     => running,
+    enable     => true,
+    hasrestart => true,
+    restart    => '/usr/bin/systemctl reload nftables',
+  }
+
+  systemd::dropin_file{'puppet_nft.conf':
+    ensure => present,
+    unit   => 'nftables.service',
+    source => 'puppet:///modules/nftables/systemd/puppet_nft.conf',
+    notify => Service['nftables'],
   }
 
   service{'firewalld':
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -5,3 +5,4 @@ fixtures:
   forge_modules:
     concat: "puppetlabs/concat"
     stdlib: "puppetlabs/stdlib"
+    systemd: "camptocamp/systemd"
--- a/README.md
+++ b/README.md
@@ -42,6 +42,14 @@ for all masterchains.
 INPUT and OUTPUT to the loopback device is allowed by
 default, though you could restrict it later.
 
+### Rules Validation
+Initially puppet deploys all configuration to
+`/etc/nftables/puppet-preflight/` and
+`/etc/nftables/puppet-preflight.nft`. This is validated with
+`nfc -c -L /etc/nftables/puppet-preflight/ -f /etc/nftables/puppet-preflight.nft`.
+If and only if successful the configuration will be copied to
+the real locations before the service is reloaded.
+
 ### nftables::config
 
 Manages a raw file in `/etc/nftables/puppet/${name}.nft`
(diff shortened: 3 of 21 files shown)


traylenator:
I'm not very proud of that `restart` line so happy to take suggestions.

traylenator:
Still thinking about this so draft for now.

duritong:
Yeah, this is generally an issue with Puppet, since Puppet forgets that it wasn't able to reload the firewall the run before and since nothing changes in the next run there is no need to re-run the reload it will never be done.

We also have this issue in the shorewall module and what we did there, was to periodically run a `shorewall check` as cron and notify whether it succeeded or not. I don't think this is something we should re-implement in the nftables module.

Also I would not like to reload the tables on each run.

But since nft supports a check run:

```
nft -h
[...]
-c, --check			Check commands validity without actually applying the changes.
```

I would suggest to trigger a nft check on any change before trying to restart the nftables service OR when the current running version does not match the current running version.

The latter could be done through concat putting things down into a *temporary* file (e.g. `/etc/nftables/puppet.nft.stage`), that is then checked, and once successfully copied (through file resource) to `/etc/nftables/puppet.nft`) and when this one is updated the service is restarted.

manifests/init.pp
```diff
@@ -71,7 +71,7 @@
     ensure     => running,
     enable     => true,
     hasrestart => true,
-    restart    => '/usr/bin/systemctl reload nftables || /usr/bin/echo "#CONFIG BROKEN" >> /etc/nftables/puppet.nft && /bin/false',
+    restart    => '/usr/bin/systemctl reload nftables || ( /usr/bin/echo "#CONFIG BROKEN" >> /etc/nftables/puppet.nft && /bin/false)',
```
duritong, line 74: maybe you need to wrapt it with `bash -c` since puppet might not run this as one to one in a shell

duritong:
But I also like your idea, as it would change the file, so we would ensure to try to update it again...

traylenator:
> Also I would not like to reload the tables on each run.

Just to be clear this will not reload every time, only when the config is changed or  broken as that always triggers a change.
Very possible that is exactly what you were saying.

Thanks for the other comments, will digest.

traylenator:
Adding the intermediate file to do a validation test before service. I will add the #CONFIG BROKEN in that file so we do still get the multiple attempts every puppet run.

manifests/init.pp
traylenator: This is correct now, it was very wrong logic before but now good.

traylenator:
So now.

1. configuration is deployed to */etc/nftables/puppet-flightest.nft* and */etc/nftables/puppet/flighttest/*
2. `/usr/sbin/nft -I /etc/nftables/puppet-preflight -c -f /etc/nftables/puppet-preflight.nft  .... `
3. Copy same configuration to */etc/nftables/puppet.nft* and */etc/nftables/puppet*
4. reload or start on that.

The result is okay. , if I add some rubbish config:

```
Error: /Stage[main]/Nftables/Exec[nft validate]: Failed to call refresh: '/usr/sbin/nft -I /etc/nftables/puppet-preflight -c -f /etc/nftables/puppet-preflight.nft || ( /usr
/bin/echo "#CONFIG BROKEN" >> /etc/nftables/puppet-preflight.nft && /bin/false)' returned 1 instead of one of [0]                                                           
Error: /Stage[main]/Nftables/Exec[nft validate]: '/usr/sbin/nft -I /etc/nftables/puppet-preflight -c -f /etc/nftables/puppet-preflight.nft || ( /usr/bin/echo "#CONFIG BROKE
N" >> /etc/nftables/puppet-preflight.nft && /bin/false)' returned 1 instead of one of [0]                                                                                   
Notice: /Stage[main]/Nftables/File[/etc/nftables/puppet.nft]: Dependency Exec[nft validate] has failures: true                   
```
and 2nd puppet run is exactly the same till fixed.

The particular awkward ones are the `includes` called inside `/etc/nftables/puppet`

It's all a bit convoluted.

The only real advantage of the 3rd commit over the 2nd commit is it covers the reboot case with broken rules.

duritong (approved):

duritong:
Excellent, this looks great!

keachi (approved):
lgtm

traylenator:
One more commit I think.
I don't like that if you run

```
nft -f /etc/nftables/puppet.nft
```
It does not fail but loads a completely wrong rule set.
Will make puppet.nft contain relative includes also so the above command fails and you have to specify the `-I` parameter.

duritong:
> It does not fail but loads a completely wrong rule set.
> Will make puppet.nft contain relative includes also so the above command fails and you have to specify the `-I` parameter.

Yeah, this could be confusing (if not even dangerous), but if we can have a workaround: even better.

traylenator:
```
nft -c -f /etc/nftables/puppet.nft 
```
now fails hard.
```
/etc/nftables/puppet.nft:11:1-26: Error: File not found: inet-filter.nft
include "inet-filter.nft"
^^^^^^^^^^^^^^^^^^^^^^^^^^
/etc/nftables/puppet.nft:13:1-22: Error: File not found: ip6-nat.nft
include "ip6-nat.nft"
^^^^^^^^^^^^^^^^^^^^^^
```
and there are comments in the file about that you need to run:

```
nft -c -I /etc/nftables/puppet -f /etc/nftables/puppet.nft 
```
to get the desired result

All rebased.

This migrates better on puppet6 than on puppet5 as the `systemctl daemon-reload` is not called early enough for the service reload.
We are puppet 6 from yesterday.

duritong:
great stuff!

PR #13, opened 2020-11-19
Add comments for all the nftable::rules entries

traylenator:
For each nftable::rule this adds an extra concat fragment to
add a comment containing the name and order number for the rule.

The motivation here is to make the mapping from resulting rules back
to puppet code more obvious. When adding a new rule it should be more
obvious to understand what order to choose.

An example resulting table ends up reading as:

```
# Start of fragment order:00 default_in header
chain default_in {
#   Start of fragment order:50 rulename:default_in-dhcpv6_client
  ip6 saddr fe80::/10 ip6 daddr fe80::/10 udp sport 547 udp dport 546 accept
#   Start of fragment order:50 rulename:default_in-ssh
  tcp dport {22} accept
#   Start of fragment order:90 rulename:default_in-drop_broadcasts
  meta pkttype broadcast counter drop
# Start of fragment order:99 default_in footer
}

```

In addition there is a new test `nftables::rule`. This includes
a pending test since I would assume setting source and content
on a rule should be an error however this currently not the case.

### DIFF
--- a/manifests/rule.pp
+++ b/manifests/rule.pp
@@ -25,9 +25,15 @@ define nftables::rule(
       $fragment = "nftables-${table}-chain-${data[0]}-rule-${data[1]}"
     }
 
+    concat::fragment{"${fragment}_header":
+      content => "#   Start of fragment order:${order} rulename:${rulename}",
+      order   => "${order}${fragment}a",
+      target  => "nftables-${table}-chain-${data[0]}",
+    }
+
     concat::fragment{
       $fragment:
-        order  => $order,
+        order  => "${order}${fragment}b",
         target => "nftables-${table}-chain-${data[0]}",
     }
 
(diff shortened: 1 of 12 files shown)


traylenator:
To add I did consider using the comment 'rule name order' within the rules but it is less obvious what to do with multline contents or `source =>` specifying rules.

manifests/rule.pp
```diff
@@ -25,9 +25,15 @@
...
+    concat::fragment{"${fragment}_header":
+      content => "#   Start of fragment order:${order} rulename:${rulename}",
+      order   => "${order}${fragment}a",
+      target  => "nftables-${table}-chain-${data[0]}",
+    }
+
     concat::fragment{
       $fragment:
-        order  => $order,
+        order  => "${order}${fragment}b",
```
duritong, line 36: Can't we do `${order}-${fragment}-(a|b)` ? this would improve readability a bit

duritong (changes requested):

nbarrientos:
> Happy to switch to prefixing rules with comments rather than extra fragments but the input for content might have to be stricter.

An advantage of using built-in comments is that they show up when executing `nft list ruleset` which might be handy.

duritong (approved):

PR #15, opened 2020-11-20
Correct bad merge

traylenator:
There was a bad merge between

* https://github.com/duritong/puppet-nftables/pull/13
* https://github.com/duritong/puppet-nftables/pull/14

correct tests.


traylenator:
More over I think a lot of these tests could be removed. e.g.


```ruby
it {
          is_expected.to contain_concat__fragment('nftables-inet-filter-chain-INPUT-rule-drop_invalid').with(
            target:  'nftables-inet-filter-chain-INPUT',
            content: %r{^  ct state invalid drop$},
            order:   '06',
            order:   '06-nftables-inet-filter-chain-INPUT-rule-drop_invalid-b',
          )
        }
```
being tested in `inet_filter_spec.rb` is I would say redundant now that `nftables::rules` is well covered by its own test.

Changing it for 

```ruby
it { is_expected to conain_nftables__rule('INPUT-drop_invalid).with(.... ) }
```

is good enough and easier to maintain moving forward.

PR #16, opened 2020-11-20
Move ICMP stuff to separate classes allowing better customisation

### LINKED ISSUE #9: The global chain contains INPUT specific rules
The current `global` chain is documented as "something we want for all" and it is used in the 3 master chains (`INPUT`, `OUTPUT` and `FORWARD`).

However, it contains rules that are specific to the `INPUT` chain, for instance:
```
ip protocol icmp icmp type echo-request limit rate 4/second accept
```

It does not really make sense to rate limit the echo request in the `OUTPUT` or `FORWARD` chain. So this rule is more for the `INPUT` chain.

IMHO, only the rules which are always valid for all the master chains should end up in the `global` chain. And I wonder whether state tracking (`ct state established,related accept`) makes sense in the `FORWARD` chain...

In any case, ICMP handling (both as input and as output) should IMHO be moved into specific rules like we have for most protocols.

nbarrientos:
This is a proposal to finish off #9. The patch moves all ICMP traffic management away from the `global` chain, creating specific input and output rules. RHEL8's defaults are followed. 

Users can customise the ICMP4/6 types to allow by providing values via Hiera to `nftables::rules::icmp::v4_types` and other similar parameters. One rule per type is added so it should be possible to add rates as part of the type, for instance:

```yaml
 nftables::rules::icmp::v4_types:
  - "echo-request limit rate 4/second"
```

ICMP rule handling can completely be disabled by tweaking `nftables::in_icmp` and `nftables::out_icmp`.

~~Tests will be added if the proposed interface goes forward.~~

Closes #9

### DIFF
--- a/files/config/puppet-inet-filter.nft
+++ b/files/config/puppet-inet-filter.nft
@@ -2,8 +2,4 @@
 
   # something we want for all
   chain global {
-    ip protocol icmp icmp type { destination-unreachable, time-exceeded, parameter-problem } accept
-    ip6 nexthdr ipv6-icmp icmpv6 type { destination-unreachable, packet-too-big, time-exceeded, parameter-problem, mld-listener-query, mld-listener-report, mld-listener-done, nd-router-advert, nd-neighbor-solicit, nd-neighbor-advert, ind-neighbor-solicit, ind-neighbor-advert, mld2-listener-report } accept
-    ip protocol icmp icmp type echo-request limit rate 4/second accept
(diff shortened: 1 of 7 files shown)


LionelCons:
API LGTM.

nbarrientos:
Tests added (won't pass due to #15 blocking it.

nbarrientos:
Rebased, tests green and ready to be reviewed. Thanks!

keachi (approved):
lgtm

duritong (approved):

PR #20, opened 2020-11-24
Make masking Service['firewalld'] optional

nbarrientos:
Users might be completely removing `firewalld` from the system when using this module. Unfortunately masking an unexisting systemd unit leads to errors when applying the catalog:

```
Info: Applying configuration version '1606207933'
Error: Could not disable firewalld:
Error: /Stage[main]/Nftables/Service[firewalld]/enable: change from 'false' to 'mask' failed: Could not disable firewalld:
```

This patch allows module users to disable the unit masking which is enough to get a clean Puppet run if there's no trace of firewalld in the system.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -42,6 +42,11 @@
 #   Adds INPUT and OUTPUT rules to allow traffic that's part of an
 #   established connection and also to drop invalid packets.
 #
+# @param firewalld_enable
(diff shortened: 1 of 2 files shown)


traylenator:
LGTM

The fact that `enable => false` is okay on a non-existent service and `enable => mask` is not okay is a puppet bug I would say.

nbarrientos:
Ready to be reviewed.

PR #22, opened 2020-11-24
Set a customisable rate limit to the logging rules

### LINKED ISSUE #19: Add an option to limit the rate of logged messages
The default logging of incoming packets comes from:
```
    'INPUT-log_discarded':
      order   => '97',
      content => "log prefix \"${sprintf($_log_prefix_discard, { 'chain' => 'INPUT' })}\" flags all counter";
```

This contains no rate limiting so this opens the door to DoS attacks.

All the rules that log something should use rate limiting, at least via an option.

nbarrientos:
As discussed in #19, this patch adds an extra variable to the main interface allowing to customise the rate limiting policy applied to the rules that log discarded traffic. By default, a limit is configured as follows:

```
chain INPUT {
...
limit rate 3/minute log prefix "[nftables] FORWARD Rejected: " flags all counter packets 0 bytes 0
...
}
# same for OUTPUT and FORWARD
```

Closes #19.

### DIFF
--- a/manifests/inet_filter.pp
+++ b/manifests/inet_filter.pp
@@ -1,7 +1,12 @@
 # manage basic chains in table inet filter
 class nftables::inet_filter inherits nftables {
 
-  $_log_prefix_discard = sprintf($nftables::log_prefix, { 'chain' => '%<chain>s', 'comment' => 'Rejected: ' })
+  $_reject_rule = epp('nftables/reject_rule.epp',
+    {
+      'log_prefix' => sprintf($nftables::log_prefix, { 'chain' => '%<chain>s', 'comment' => 'Rejected: ' }),
+      'log_limit'  => $nftables::log_limit
+    }
+  )
 
   nftables::config{
     'inet-filter':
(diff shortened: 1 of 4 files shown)


manifests/inet_filter.pp
```diff
@@ -1,7 +1,19 @@
...
 
-  $_log_prefix_discard = sprintf($nftables::log_prefix, { 'chain' => '%<chain>s', 'comment' => 'Rejected: ' })
+  $_reject_rule_template = @(END)
+<% if $log_limit { -%>
+limit rate <%= $log_limit %> log prefix "<%= $log_prefix %>" flags all counter
+<% } else { -%>
+log prefix "<%= $log_prefix %>" flags all counter
+<% } -%>
+END
+  $_reject_rule = inline_epp($_reject_rule_template,
```
duritong, line 11: can we move that to an external template, since this really clutters the readability of the class

duritong (changes requested):
except for the template LGTM

manifests/inet_filter.pp
nbarrientos: No problem, doing :)

duritong (approved):

duritong:
@keachi PTAL

keachi (approved):
lgtm

PR #25, opened 2020-11-26
Allow disabling default NAT tables and chains

### LINKED ISSUE #23: Add an option to disable NAT
Most of what is instantiated by default can be disabled with the `$in_*` and `$out_*` variables. This is good and allows for a clean and minimal `nftables` configuration.

Unfortunately, the NAT related bits cannot be disabled. This creates two tables and four chains which are not used (by default).

It would be cleaner to have an option to remove all the NAT related code. `init.pp` could simply have something like:
```
if ($nat) {
    include nftables::ip_nat
}
```

nbarrientos:
This patch adds a parameter to the main interface to prevent the module from including `nftables::ip_nat` and hence from configuring any of the default tables and chains dedicated to processing NATed traffic.

Closes #23.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -32,6 +32,9 @@
 # @param in_icmp
 #   Allow inbound ICMPv4/v6 traffic.
 #
+# @param nat
+#   Add default tables and chains to process NAT traffic.
+#
 # @param log_prefix
 #   String that will be used as prefix when logging packets. It can contain
 #   two variables using standard sprintf() string-formatting:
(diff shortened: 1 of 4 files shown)


traylenator:
This will won't merge well with #10  but LGTM

duritong:
if you can rebase, would be great.

nbarrientos:
Rebased and conflicts resolved. Thanks for reviewing.

duritong (approved):

PR #28, opened 2020-11-27
Do not test nftables::rules repeatadly

traylenator:
Rather than testing the contents of nftable::rules just test
that nftables::rules instance is correct.

The existing test for the define nftables::rules is enough.

Motivation here is to make changes to nftables::rules easier to handle
down the line.

### DIFF
--- a/spec/classes/bridges_spec.rb
+++ b/spec/classes/bridges_spec.rb
@@ -5,49 +5,37 @@ describe 'nftables' do
 
   on_supported_os.each do |os, os_facts|
     context "on #{os}" do
-      let(:facts) { os_facts }
+      context 'with bridge interfaces br0 and br1-2' do
+        let(:facts) do
+          os_facts.merge(
+            networking:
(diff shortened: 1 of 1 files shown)


traylenator:
These tests pass but I am confused why.  Shouldn't the nftables::rule for the bridge `br1-2` be.

```
it { is_expected.to contain_nftables__rule('default_fwd-bridge_br1_2_br1_2') }
```

I'm missing something.

If this is accepted I'll repeat this execise with other cases where nftables::rules and ::chains are being checked
multiple times.

duritong (approved):
lgtm

traylenator:
Why is the passing test
```
it { is_expected.to contain_nftables__rule('default_fwd-bridge_br1_br1') }
```
and not
```
it { is_expected.to contain_nftables__rule('default_fwd-bridge_br1_2_br1_2') }
```

keachi (approved):
lgtm

PR #30, opened 2020-11-30
Scientific Linux 8 will never exist

traylenator:
As per

https://listserv.fnal.gov/scripts/wa.exe?A2=ind1904&L=SCIENTIFIC-LINUX-ANNOUNCE&P=78


traylenator:
May have spoken too soon here....

PR #31, opened 2020-11-30
New parameter noflush_tables to selectivly skip flush

traylenator:
Introduces a new structured fact nftables

```yaml
nftables:
  tables:
    - inet-filter
    - ip-nat
    - ip6-nat
    - inet-f2b-table
```

By default the nft script will continue to contain `nft flush ruleset`

If noflush_tables is specified e.g.

```puppet
class{nftables:
  noflush_tables => ['inet-f2b-table'],
}
```
the results script will explicity flush the other tables only.
i.e.

```
flush table inet filter
flush table ip nat
flush table ip6 nat
```

Motivation here is to allow a maintence of chain to managed by something else.
This example for fail2ban but could be docker, ...

### DIFF
--- a/files/systemd/puppet_nft.conf
+++ b/files/systemd/puppet_nft.conf
@@ -1,7 +1,0 @@
-# Specify directory to look for relative includes
-[Service]
-ExecStart=
-ExecStart=/sbin/nft -I /etc/nftables/puppet -f /etc/sysconfig/nftables.conf
-ExecReload=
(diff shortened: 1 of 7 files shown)


traylenator:
We still have that duplicate flush in nft script and in the systemd unit.
If I was going to drop one I would do it in systemd unit.

traylenator:
tests fail in ruby 2.4.5 only :-(

duritong:
`{ a: 'foo' } == { :a => 'foo' }`

traylenator:
Yes have revised by ruby hashes now.

Just surprising the the wrong thing worked with new puppet and ruby.

Rebasing.

PR #33, opened 2020-12-03
Add nftables::simplerule

### LINKED ISSUE #17: Rule API
This is not to report a bug but to trigger a discussion on the rule API.

Looking at `http` versus `ssh`, we see that the API is not consistent. With `http`, one can only allow a single http server running on a fixed port. With `ssh`, one can allow multiple ssh servers on any ports (thanks to the `ports` parameter). It would probably be good to have a common set of parameters (like the one to change the port) for all the service specific rules. Note: `ssh` currently uses `ports` while `masquerade` uses `dport`.

Also, `nftables::rules::http` is a class while `nftables::rule` is a defined type. So one cannot instantiate `nftables::rules::http` multiple times, as it would be natural to do to declare multiple http servers.

So, changing to a defined type and adding parameters, one could declare multiple http servers this way:
```
nftables::rules::http { 'foo':
    # note: port not needed as we use the default
    comment => 'foo web server',
}
nftables::rules::http { 'bar':
    port    => 8000,
    comment => 'bar web server',
}
```

When using `nftables` features like comments or counter or logging (i.e. `log prefix`), it could make sense to have multiple individual rules instead of a single rule with multiple ports. This would provide more flexibility.

Comments?

nbarrientos:
Add a `Firewall`-like interface covering some simple cases to ease migration of rules.

Closes #17.


LionelCons:
Good start!

My initial comments (subject to discussion) are the following.

I do not see where `$setname` is used.

There should be both `dport` and `sport` parameters. They should accept an integer or an array of integers.

Similarly, there should be `saddr` and `daddr` parameters accepting a string or an array of strings.

There is the question of IPv4 versus IPv6. I wonder whether the `proto` should be for instance `tcp` (= TCP for IPv4 and IPv6) or `tcp4` (= TCP for IPv4) or `tcp6` (= TCP for IPv6). If not, we might need a way to differentiate `ip` and `ip6`.

nbarrientos:
> I do not see where $setname is used.

Yup, it's a leftover (I used `nftables::set` as starting point). It's fixed now, thanks.

> There should be both dport and sport parameters. They should accept an integer or an array of integers.

I've implemented a richer `dport`, allowing a single port, a range and an array of ports. The same could be done for `sport`.

> Similarly, there should be saddr and daddr parameters accepting a string or an array of strings.

This is tricky as the matcher for the rule would be different if the address (or the range in CIDR notation) is an IPv4 or an IPv6 one. It could be a single parameter of type `Array[Variant[Stdlib::IP::Address::V6, Stdlib::IP::Address::V4]]` and then in the template depending on the type of the element we do `ip` or `ip6` in the statement.

LionelCons:
Thanks for these changes.

For ports, please note that there is also the question of port names. Do we allow `port=>'http'`? If yes then it is more flexible but there will be less data validation (I mean more invalid data allowed).

For addresses, there is a similar question: do we allow set names like in `saddr=>'@trusted-ips'`?

In this tradeoff between flexibility and data validation I don't know where to put the cursor...

nbarrientos:
I've just added a proposed implementation for `daddr` and the `counter` bit. I won't add the dual `saddr` and `sport` until we've reached a satisfactory implementation their counterparts. I'll stop here for the time being, lemme know what you think.

nbarrientos:
> For ports, please note that there is also the question of port names. Do we allow port=>'http'? If yes then it is more flexible but there will be less data validation (I mean more invalid data allowed).

Yes, it makes sense, but what about `dport => ['http', 'https']`? That should be legal too.

traylenator:
> Yes, it makes sense, but what about dport => ['http', 'https']? That should be legal too.

No.  Espeically since you are permitted numerics in service names the validation loss is too great and your never know what is in the `/etc/services` file on the destination file exactly.

duritong:
I like it so far

nbarrientos:
So, shall I implement `saddr` and `sport` and document the interface? What do we finally do about the destination table? What about `iifname` and `oifname`?

templates/simplerule.epp
```diff
@@ -0,0 +1,55 @@
...
+  }
+} else {
+  $_dst_hosts = undef
+} -%>
+<%- if $proto and $dport {
+  if $dport =~ Array {
+    $_dst_port = "${_proto} dport {${dport.join(', ')}}"
+  } else {
+    $_dst_port = "${_proto} dport $dport"
+  }
```
ekohl, line 41: Just a suggestion for the future: I've seen this pattern in a few places. Perhaps a separate function should be introduced to format an array/string as nft accepts it.

duritong: or we could make sure dport is always an array meaning to wrap it as an array and then always use `{${dport.join(', ')}}`

But pay attention since dport can be integers you can't do a simple `Array($dport)` as this has a different semantic than `any2array($dport)` from stdlib.

so `$_dst_port = "${_proto} dport {${any2array($dport).join(', ')}}"`

nbarrientos: Yes, we could keep `Optional[Variant[Array[Stdlib::Port, 1], Stdlib::Port, Pattern[/\d+-\d+/]]]` as type and treat it always as an array.

> But pay attention since dport can be integers you can't do a simple Array($dport) as this has a different semantic than any2array($dport) from stdlib.

This should do the job using `Array()`, note the `true` ([docs](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/REFERENCE.md#any2array-1)):

```
$_dst_port = "${_proto} dport {${Array($dport, true).join(', ')}}"
```

In action:

```puppet
notify{"tcp dport {${Array(53, true).join(', ')}}":}
notify{"tcp dport {${Array([53, 54], true).join(', ')}}":}
notify{"tcp dport {${Array('53-54', true).join(', ')}}":}
```

Output:

```
Notice: Compiled catalog for example in environment production in 0.03 seconds
Notice: tcp dport {53}
Notice: /Stage[main]/Main/Notify[tcp dport {53}]/message: defined 'message' as 'tcp dport {53}'
Notice: tcp dport {53, 54}
Notice: /Stage[main]/Main/Notify[tcp dport {53, 54}]/message: defined 'message' as 'tcp dport {53, 54}'
Notice: tcp dport {53-54}
Notice: /Stage[main]/Main/Notify[tcp dport {53-54}]/message: defined 'message' as 'tcp dport {53-54}'
Notice: Applied catalog in 0.13 seconds
```

The three above are valid syntax-wise for nftables.

traylenator:
> So, shall I implement saddr and sport and document the interface? What do we finally do about the destination table? What about iifname and oifname?

saddr and sport I would say are useful, apart from anything else it's presumably same code as daddr and dcode.

As for i/oname can leave for now and add later if requested. It would be reasnable request.

nbarrientos:
`sport` and `saddr` added. Some documentation added, too.

manifests/simplerule.pp
```diff
@@ -0,0 +1,99 @@
...
+  Enum['present','absent'] $ensure = 'present',
+  Pattern[/^[-a-zA-Z0-9_]+$/] $rulename = $title,
+  Pattern[/^\d\d$/] $order = '50',
+  String $chain  = 'default_in',
+  String $table = 'inet-filter',
+  Enum['accept', 'continue', 'drop', 'queue', 'return'] $action = 'accept',
+  Optional[String] $comment = undef,
+  Optional[Variant[Array[Stdlib::Port, 1], Stdlib::Port, Pattern[/\d+-\d+/]]] $dport = undef,
+  Optional[Enum['tcp', 'tcp4', 'tcp6', 'udp', 'udp4', 'udp6']] $proto = undef,
+  Optional[Variant[Stdlib::IP::Address::V6, Stdlib::IP::Address::V4, Pattern[/^@[-a-zA-Z0-9_]+$/]]] $daddr = undef,
```
traylenator, line 66: Given this is now duplicated and also a bit a of a monster create a new type `Nft::Ports` it can be used elsewhere as well in time.

Same for the s/daddr

nbarrientos: Sure.

types/port.pp
```diff
@@ -0,0 +1,5 @@
+type Nftables::Port = Variant[
```
traylenator, line 1: For the types you can spec tests and string docs.

duritong (approved):

duritong:
all good, but you need to rebase

traylenator (approved):

traylenator:
@LionelCons anything to add ?

LionelCons:
Yes, I have something to add...

This is excellent work, thank you very much!

PR #34, opened 2020-12-03
Remove duplicate flush on reload

traylenator:
When nftables was reloaded a flush was being done both in the systemd
reload call and in the nft script itself.


duritong (approved):

manifests/init.pp
```diff
@@ -145,11 +145,11 @@
     restart    => '/usr/bin/systemctl reload nftables',
   }
 
-  systemd::dropin_file { 'puppet_nft.conf':
-    ensure  => present,
-    unit    => 'nftables.service',
-    content => epp('nftables/systemd/puppet_nft.conf.epp', { 'noflush' => $noflush_tables }),
-    notify  => Service['nftables'],
+  systemd::dropin_file{'puppet_nft.conf':
```
duritong, line 148: this should be brought back to what lintern wants

duritong (changes requested):

duritong:
I rebased but found a minor thing to fix.

manifests/init.pp
```diff
@@ -145,11 +145,11 @@
...
+    ensure => present,
+    unit   => 'nftables.service',
+    source => 'puppet:///modules/nftables/systemd/puppet_nft.conf',
```
ekohl, line 151: I generally prefer to use `file()`. Also keeps the diff smaller. At runtime this is generally faster, unless files increase to a certain size.
```suggestion
    content => file('nftables/systemd/puppet_nft.conf'),
```

ekohl (changes requested):

manifests/init.pp
traylenator: Why have I never used `content => file(...)` before.

ekohl: I can't find the blog post now (it's already many years old), but I think that up to a few KB in size inlining it in the catalog is faster than the separate HTTP request. The added bonus is that it's part of catalog compilation. That means your unit tests catch it when the file isn't actually present rather than failing at runtime.

ekohl (approved):
Not commenting on the actual change, but :+1: from a code perspective.

duritong (approved):

manifests/init.pp
binford2k: It also means that you can enforce a cached catalog while offline. We've had some users take advantage of that for eg roaming device management.

@ekohl your premise is still accurate, though the numbers in that post are probably not valid anymore. Puppet's been through several performance optimizations since.

PR #36, opened 2020-12-07
modulesync  4.0.0

traylenator:
First modulesync of puppet-nftables since migration.

* modulesync 4.0.0
* drop original .sync.yml with `~/.gitlab-ci.yml` references
* Remove puppet 4 support
* lint fixes
* prarameter definitions on one line.


ekohl (commented):
Should we drop the PDK part as well:
https://github.com/voxpupuli/puppet-nftables/blob/e3c56ff6d70945d049583754b2b293200765f26d/metadata.json#L49-L51

manifests/bridges.pp
```diff
@@ -1,9 +1,9 @@
 # allow forwarding traffic on bridges
-class nftables::bridges(
+class nftables::bridges (
   Enum['present','absent']
-    $ensure = 'present',
+  $ensure = 'present',
```
ekohl, line 4: I think this is less readable. Perhaps the lint plugin should understand this. Personally I don't like splitting definitions over multiple lines but I can understand why you'd prefer it over aligning. Personally I stopped aligning altogether.

manifests/rules/icinga2.pp
```diff
@@ -1,9 +1,9 @@
 # manage in icinga2
-class nftables::rules::icinga2(
+class nftables::rules::icinga2 (
   Array[Integer,1]
```
ekohl, line 3: Unrelated, but should it use `Stdlib::Port` instead of `Integer`?

ekohl (commented):
Comments are applicable in many places so I just commented once.

traylenator:
In the abscence of CI testing , two test failiures.

```
  1) nftables nft rules present returns valid tables
     Failure/Error: Process.stubs(:uid).returns(0)
     
     NoMethodError:
       undefined method `stubs' for Process:Module
       Did you mean?  stub
     # ./spec/unit/facter/nftables_spec.rb:6:in `block (2 levels) in <top (required)>'
     # /usr/share/gems/gems/bundler-2.1.4/libexec/bundle:46:in `block in <top (required)>'
     # /usr/share/gems/gems/bundler-2.1.4/libexec/bundle:34:in `<top (required)>'

  2) nftables nft fails does not return a fact
     Failure/Error: Process.stubs(:uid).returns(0)
     
     NoMethodError:
       undefined method `stubs' for Process:Module
       Did you mean?  stub
     # ./spec/unit/facter/nftables_spec.rb:6:in `block (2 levels) in <top (required)>'
     # /usr/share/gems/gems/bundler-2.1.4/libexec/bundle:46:in `block in <top (required)>'
     # /usr/share/gems/gems/bundler-2.1.4/libexec/bundle:34:in `<top (required)>'
```

ekohl:
> In the abscence of CI testing , two test failiures.

The workaround is to push the branch to upstream. Then you can delete it. Github actions will continue to run after that, also on new pushes.

traylenator:
puppet 4 refuses to go, guss the matrix is what is generated on the voxpupuli namespace.

manifests/bridges.pp
traylenator: I agree both on one line and by not aligning you avoid this nasty diffs when one new longer parameter line is added.

ekohl (approved):

traylenator:
@ekohl many thanks - think I have run out of things to do in this first pass.

PR #38, opened 2020-12-07
Add basic configuration validation acceptance test

traylenator:
#### Add basic configuration validation acceptance test

It is not possible to start the nftables service within docker so
the service is altered to only validate the service
configuration resulting from concat constructed files and nft inclusions.

### DIFF
--- a/spec/acceptance/default_spec.rb
+++ b/spec/acceptance/default_spec.rb
@@ -4,7 +4,8 @@ describe 'nftables class' do
   context 'configure default nftables service' do
     it 'works idempotently with no errors' do
       pp = <<-EOS
-      # default mask of firewalld service fails if service not installed.
+      # default mask of firewalld service fails if service is not installed.
+      # https://tickets.puppetlabs.com/browse/PUP-10814
       class { 'nftables':
         firewalld_enable => false,
       }
--- a/spec/acceptance/default_spec.rb
+++ b/spec/acceptance/default_spec.rb
@@ -1,0 +1,50 @@
+require 'spec_helper_acceptance'
+
+describe 'nftables class' do
+  context 'configure default nftables service' do
+    it 'works idempotently with no errors' do
+      pp = <<-EOS
+      # default mask of firewalld service fails if service not installed.
+      class { 'nftables':
+        firewalld_enable => false,
+      }
+      # nftables cannot be started in docker so replace service with a validation only.
+      systemd::dropin_file{"zzz_docker_nft.conf":
+        ensure  => present,
+        unit    => "nftables.service",
+        content => [
+          "[Service]",
+          "ExecStart=",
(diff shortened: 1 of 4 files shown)


traylenator:
? acceptance test is not running.....

ekohl:
There is one check here:
https://github.com/voxpupuli/modulesync_config/blob/f9c50882764313a7ea8736b9d79a91d5dbcaad2b/moduleroot/.github/workflows/ci.yml.erb#L48

Perhaps that can also be expressed with some if or [puppet_metadata](https://github.com/voxpupuli/puppet_metadata) could generate an empty matrix, but we needed to migrate away from Travis quickly.

For now it means that you do need to regen modulesync once you've added acceptance tests. After that it should be dynamic whenever you add/remove some OS.

traylenator:
Timeout after 6 hours.

Guessing port 5689 that is used for SSH into container(?) needs to left open.

traylenator:
Starting nftables inside docker seems to be doomed however the nft -c validate stage works perfectly
so this is still a worth while exercise. Configure to ignore the service.

bastelfreak (approved):

nbarrientos (approved):
LGTM

spec/acceptance/default_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'spec_helper_acceptance'
+
+describe 'nftables class' do
+  context 'configure default nftables service' do
+    it 'works idempotently with no errors' do
+      pp = <<-EOS
+      # default mask of firewalld service fails if service not installed.
```
ekohl, line 7: That sounds like a bug in the module.

traylenator: It's bug in puppet possibly.

```puppet
service{'foo':
  enable => false,
}
```
is okay if service foo does not exist however with `enable => mask` it gives an error.

So you want to do something different depending upon if firewalld is installed or not but this module
does not  enforce if firewalld should be installed.

Puppet should mask anyway.

```
systemctl mask foobar.service
Unit foobar.service does not exist, proceeding anyway.
Created symlink /etc/systemd/system/foobar.service → /dev/null.
```
seems to work.

Will see if that is upstream.

traylenator: Nothing exact - https://tickets.puppetlabs.com/browse/PUP-10814 submitted.

So given the behviour of puppet mask is sensible default in case the package is installed which is normally the case but for other cases like this one it needs an option.

Resolved for me.

ekohl: I always like to add a link to the issue as a code comment because that would answer my question. Other that it looks good to me.

bastelfreak (approved):

ekohl (approved):

PR #39, opened 2020-12-08
rubocop:auto_correct fixes

traylenator:
#### rubocop corrections
Results of rubocop:auto_correct

### DIFF
--- a/spec/classes/dnat4_spec.rb
+++ b/spec/classes/dnat4_spec.rb
@@ -56,28 +56,28 @@ describe 'nftables' do
             owner:          'root',
             group:          'root',
             mode:           '0640',
-            ensure_newline: true,
+            ensure_newline: true
           )
         }
         it {
           is_expected.to contain_concat__fragment('nftables-inet-filter-chain-default_fwd-header').with(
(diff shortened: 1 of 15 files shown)


traylenator:
no rubocop CI yet.

ekohl:
Yes, that is something I recently discovered. In the migration from Travis to Github Actions I missed that.

spec/classes/dnat4_spec.rb
```diff
@@ -56,71 +56,71 @@
             owner:          'root',
             group:          'root',
             mode:           '0640',
-            ensure_newline: true,
```
ekohl, line 59: FYI: we had some discussion about trailing comma's. I think most people agreed that we should enable them in Rubocop as well. We use them in Puppet so it's inconsistent not to do the same in Ruby. I would prefer that over changing everything in this module.

ekohl (changes requested):

spec/classes/dnat4_spec.rb
traylenator: Agreed.

PR #40, opened 2020-12-08
Comment why firewalld_enable parameter is required

traylenator:
#### Comment why firewalld_enable parameter is required
service.firewalld.enable being default masked or false is due to a missing feature in puppet.
Add a comment 

https://tickets.puppetlabs.com/browse/PUP-10814

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -152,6 +152,8 @@ class nftables (
     notify  => Service['nftables'],
   }
 
+  # firewalld.enable can be mask or false depending upon if firewalld is installed or not
+  # https://tickets.puppetlabs.com/browse/PUP-10814
   service { 'firewalld':
     ensure => stopped,
     enable => $firewalld_enable,


manifests/init.pp
```diff
@@ -152,6 +152,8 @@
     notify  => Service['nftables'],
   }
 
+  # firewalld.enable can be mask or false depending upon if firewalld is installed or not
```
ekohl, line 155: Would it be better to move this to the `@param` documentation for `$firewalld_enable`?

traylenator: It's there already.

```puppet
# @param firewalld_enable
#   Configures how the firewalld systemd service unit is enabled. It might be
#   useful to set this to false if you're externaly removing firewalld from
#   the system completely.
```

 Adding the link there is a bit ugly and here in the code makes it obvious as something to review for the future.

ekohl (approved):

PR #42, opened 2020-12-08
switch not the server naming

duritong:
The world moved on to the term puppetserver

### DIFF
--- a/manifests/rules/out/puppet.pp
+++ b/manifests/rules/out/puppet.pp
@@ -1,19 +1,19 @@
 # manage outgoing puppet
 class nftables::rules::out::puppet (
-  Variant[String,Array[String,1]] $puppetmaster,
+  Variant[Stdlib::IP::Address,Array[Stdlib::IP::Address,1]] $puppetserver,
   Integer $puppetserver_port = 8140,
 ) {
-  any2array($puppetmaster).each |$index,$pm| {
+  Array($puppetserver, true).each |$index,$ps| {
     nftables::rule {
       "default_out-puppet-${index}":
     }
-    if $pm =~ /:/ {
+    if $ps =~ Stdlib::IP::Address::V6 {
       Nftables::Rule["default_out-puppet-${index}"] {
-        content => "ip6 daddr ${pm} tcp dport ${puppetserver_port} accept",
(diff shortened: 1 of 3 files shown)


ekohl:
I don't see a backwards-incompatible label. @bastelfreak I thought the app was supposed to sync those.

bastelfreak:
@ekohl added it. the app is currently in maintenance mode because the github api trolled me a few times. I hope to fix this next week.

manifests/rules/out/puppet.pp
```diff
@@ -1,19 +1,19 @@
 # manage outgoing puppet
 class nftables::rules::out::puppet (
-  Variant[String,Array[String,1]] $puppetmaster,
+  Variant[String,Array[String,1]] $puppetserver,
   Integer $puppetserver_port = 8140,
 ) {
-  any2array($puppetmaster).each |$index,$pm| {
+  any2array($puppetserver).each |$index,$ps| {
```
bastelfreak, line 6:
```suggestion
  [$puppetserver].flatten.each |$index,$ps| {
```

this should work as well and avoids the stdlib function call

bastelfreak (approved):

manifests/rules/out/puppet.pp
```diff
@@ -1,19 +1,19 @@
 # manage outgoing puppet
 class nftables::rules::out::puppet (
```
traylenator, line 2: There's no tests for this class, maybe be add some trivial ones.

traylenator (changes requested):

manifests/rules/out/puppet.pp
duritong: yeah wasn't sure, whether this should happen in this change

traylenator:
The new test spec/acceptance/all_rules_spec.rb in #50 will need updating depending on if this or #50 gets there first - trivial.

nbarrientos:
Needs rebasing to make the acceptance tests fail.

manifests/rules/out/puppet.pp
duritong: added

traylenator (approved):

PR #46, opened 2020-12-09
WIP - fix #45 - add custom_file define

### LINKED ISSUE #45: deploying custom tables is broken
You can't anymore deploy a custom config using nftables::config

We do

```
include "custom-*.nft"
```

But there is no way how you can create such a file with the reworked nftables::config, since the name is used for the filter table afterwards.

This got broken in e4c322229612b708ab6e07baf719ca4bd2f61d91

CC @nbarrientos 

I am trying to add a custom ip table for fail2ban: based on: https://wiki.meurisse.org/wiki/Fail2Ban and I don't think we can do it anymore with the current state of the module.

duritong:
This is a WIP for the issue in #49 and how I am now deploying custom configs.

We could even hardcode the `custom-` in front.

Once the design is clear, I could add tests and fix documentation.

### DIFF
--- a/manifests/custom_file.pp
+++ b/manifests/custom_file.pp
@@ -1,0 +1,29 @@
+# manage a custom config file as it is
+define nftables::custom_file (
+  Optional[String] $content = undef,
+  Optional[Variant[String,Array[String,1]]] $source = undef,
+) {
+  Package['nftables'] -> file {
+    "/etc/nftables/puppet-preflight/${name}.nft":
+      owner => root,
+      group => root,
+      mode  => '0640',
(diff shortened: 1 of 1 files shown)


traylenator:
One think that could simplify all these individual file copies is we could just use a single file to copy the whole directory from preflight to live directory. Would remove a load of code.

duritong:
> One think that could simplify all these individual file copies is we could just use a single `file` to copy the whole directory from preflight to live directory. Would remove a load of code.

Oh that's actually an excellent idea and it would also purge things...

duritong:
We have a successor #48

PR #48, opened 2020-12-10
Several fixes for nftables::config

nbarrientos:
A few improvements to `nftables::config`, derived from #45.

* Autogenerate the statement to include chains to save templates for the simple cases
* Actually fail if both `content` and `source` are passed (this was a TODO, I believe)
* Remove duplicate context name
* Kind of validate the resource title so all tests pass

This patch complements (and might conflict with) #47

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -1,5 +1,6 @@
 # manage a config snippet
 define nftables::config (
+  Pattern[/^\w+-\w+$/] $tablespec = $title,
   Optional[String] $content = undef,
   Optional[Variant[String,Array[String,1]]] $source = undef,
 ) {
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -38,11 +38,16 @@ define nftables::config (
         source => $source,
     }
   } else {
+    if $content {
+      $_content = $content
+    } else {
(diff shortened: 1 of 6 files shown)


manifests/config.pp
```diff
@@ -1,8 +1,13 @@
 # manage a config snippet
 define nftables::config (
+  Pattern[/^\w+-\w+$/] $tablespec = $title,
   Optional[String] $content = undef,
   Optional[Variant[String,Array[String,1]]] $source = undef,
 ) {
+  if $content and $source {
+    fail('Please pass only $content or $source, not both.')
```
duritong, line 8: this does not check the case, when none is passed.

duritong (changes requested):
except for the additional check, all good for me.

manifests/config.pp
nbarrientos: Uh, sorry, I'm not sure what you mean. Passing neither is fine, that's the main goal of the change. To allow just declaring tables like this:

```puppet
nftables::config{'ip-fail2ban':}
```

but guess that's not what you mean. Could you please clarify?

duritong: ahhh, right :p well I already forgot about that when I reviewed this.

duritong (approved):

PR #50, opened 2020-12-10
Correct nfs3 invalid udp /tcp matching rule and more tests

traylenator:
#### Correct invalid nfs3 udp/tcp matching rule
The udp /tcp matching rule for nfs3 was missing a `th` apparently as the per the examples.

In addition more acceptance tests:

* It should be possible to include all rules
* If a bad invalid rule for nft verification is added the real nft configuration should be left in a good state.

### DIFF
--- a/manifests/rules/nfs3.pp
+++ b/manifests/rules/nfs3.pp
@@ -2,6 +2,6 @@
 class nftables::rules::nfs3 {
   nftables::rule {
     'default_in-nfs3':
-      content => 'meta l4proto { tcp, udp } dport nfs accept comment "Accept NFS3"',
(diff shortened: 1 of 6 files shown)


traylenator:
@dvanders slight correction.

nbarrientos:
As a thought for the future, it might make sense to add a CI check to make sure that all the predefined rules are represented in this acceptance test so new ones are not left behind.

nbarrientos (approved):

traylenator:
> As a thought for the future, it might make sense to add a CI check to make sure that all the predefined rules are represented in this acceptance test so new ones are not left behind.

Any thought how it might be done other than glob'ing the filesystem?

duritong (approved):

PR #51, opened 2020-12-10
Enable Puppet 7 support

bastelfreak:


traylenator (approved):
I've yet to find anything that does not work with 7.

PR #52, opened 2020-12-10
Remove a blank separating the doc string and the code

nbarrientos:
Otherwise the generator of the docs does not do the job :/

Also shorten the description to comply with the rules. The recommendation is already in the README.

### DIFF
--- a/manifests/simplerule.pp
+++ b/manifests/simplerule.pp
@@ -1,6 +1,4 @@
-# @summary Provides a simplified interface to nftables::rule for basic use cases.
-#   It's recommended to use nftables::rule directly if you feel comfortable with
-#   nft's syntax.
+# @summary Provides a simplified interface to nftables::rule
 #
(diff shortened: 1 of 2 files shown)


manifests/simplerule.pp
```diff
@@ -1,6 +1,4 @@
-# @summary Provides a simplified interface to nftables::rule for basic use cases.
-#   It's recommended to use nftables::rule directly if you feel comfortable with
-#   nft's syntax.
+# @summary Provides a simplified interface to nftables::rule
```
traylenator, line 1: Just for info you can do multiline summaries with:
```
# @summary
#  Line 1
#  Line 2
```

traylenator (approved):

ekohl:
FYI: https://github.com/rodjek/puppet-lint/commit/cb2b40e469ca019911d8980e9cc1f53191a89de7 is still pending a release, but that would have caught this.

PR #53, opened 2020-12-11
Check that all the predefined rules are declared in the all rules acceptance test

nbarrientos:
Can be seen in action here: https://github.com/voxpupuli/puppet-nftables/pull/54/checks?check_run_id=1537808159

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,8 @@ jobs:
         with:
           ruby-version: '2.7'
           bundler-cache: true
+      - name: Check that all rules are declared in the all rules acceptance test
+        run: bash -c "awk '/^class / {print \$2}' manifests/rules/*.pp manifests/rules/out/*.pp | xargs -i sh -c 'grep -F {} spec/acceptance/all_rules_spec.rb -q || exit 255'"
       - name: Run tests
         run: bundle exec rake beaker
         env:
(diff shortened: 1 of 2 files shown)


traylenator (commented):
This will be overwritten by modulesync.

Looking at the template https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.github/workflows/ci.yml.erb

There is no parameter to add extra tests which there used to be with the old Travis templates I think. (Maybe confusing with our own templates).

traylenator:
> This will be overwritten by modulesync.
> 
> Looking at the template https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.github/workflows/ci.yml.erb
> 
> There is no parameter to add extra tests which there used to be with the old Travis templates I think. (Maybe confusing with our own templates).
> 
Will try and extend the templates.

traylenator (approved):

nbarrientos:
🤷

.github/workflows/ci.yml
```diff
@@ -70,6 +70,8 @@ jobs:
         with:
           ruby-version: '2.7'
           bundler-cache: true
+      - name: Check that all rules are declared in the all rules acceptance test
```
traylenator, line 73: Was bit quick to approve here , given this will be overwritten I assume adding a whole new file in this directory will be okay. Maybe that's the best thing to do.

traylenator (changes requested):
Do this in away that it will not be overwritten.

traylenator (approved):

PR #59, opened 2021-01-13
Align simplerule and rule rulename requirements

### LINKED ISSUE #58: nftables::simplerule should follow the same rules as nftables::rule

```
  nftables::simplerule{'thanos_gRPC-store':
    action => 'accept',
    dport  => 161,
    proto  => 'udp6',
    saddr  => '2001:1458:d00:3b::100:23b',
  }
```

## What are you seeing

```
Error while evaluating a Resource Statement, Nftables::Rule[default_in-thanos_gRPC-store]: parameter 'rulename' expects a match for Pattern[/^[a-zA-Z0-9_]+-[a-zA-Z0-9_]+(-\d+)?$/], got 'default_in-thanos_gRPC-store'
```

Compilation faillure, but at `nftables::simplerule` level.

This regex:

https://github.com/duritong/puppet-nftables/blob/master/manifests/simplerule.pp#L57

Should be a mirror of the second part of this one:

https://github.com/duritong/puppet-nftables/blob/master/manifests/rule.pp#L6

Trivial fix. Add a test to cover a possible regression.

nbarrientos:
This patchsets unifies the regular expression that validates the underlying rule name in `nftables::rule` to the expression used in `nftables::simplerule` so it's clearer for the user what's going on if a bad name is used. The caller using `nftables::simplerule` does not necessarily have to know anything about `nftables::rule`.

Fixes #58

### DIFF
--- a/manifests/simplerule.pp
+++ b/manifests/simplerule.pp
@@ -54,7 +54,7 @@
 #   Enable traffic counters for the matched traffic.
 define nftables::simplerule (
   Enum['present','absent'] $ensure = 'present',
-  Pattern[/^[-a-zA-Z0-9_]+$/] $rulename = $title,
+  Pattern[/^[a-zA-Z0-9_]+(-\d+)?$/] $rulename = $title,
   Pattern[/^\d\d$/] $order = '50',
   String $chain  = 'default_in',
   String $table = 'inet-filter',
--- a/manifests/simplerule.pp
+++ b/manifests/simplerule.pp
@@ -54,7 +54,7 @@
 #   Enable traffic counters for the matched traffic.
 define nftables::simplerule (
   Enum['present','absent'] $ensure = 'present',
-  Pattern[/^[a-zA-Z0-9_]+(-\d+)?$/] $rulename = $title,
+  Nftables::SimpleRuleName $rulename = $title,
   Pattern[/^\d\d$/] $order = '50',
   String $chain  = 'default_in',
   String $table = 'inet-filter',
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1286,7 +1286,7 @@ Default value: `'present'`
 
 ##### `rulename`
 
-Data type: `Pattern[/^[-a-zA-Z0-9_]+$/]`
+Data type: `Pattern[/^[a-zA-Z0-9_]+(-\d+)?$/]`
 
 The symbolic name for the rule to add. Defaults to the resource's title.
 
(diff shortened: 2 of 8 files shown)


manifests/simplerule.pp
```diff
@@ -54,7 +54,7 @@
 #   Enable traffic counters for the matched traffic.
 define nftables::simplerule (
   Enum['present','absent'] $ensure = 'present',
-  Pattern[/^[-a-zA-Z0-9_]+$/] $rulename = $title,
+  Pattern[/^[a-zA-Z0-9_]+(-\d+)?$/] $rulename = $title,
```
traylenator, line 57: How about creating a `Nftables::Rulename`  type  ? That regex is probably elsewhere as well - not looked.

Should stop it diverging.

nbarrientos: The [regex used in `nftables::simplerule`](https://github.com/voxpupuli/puppet-nftables/blob/6a4ffeadf03a0e25d32963d12dcfa68a1a1595d0/manifests/simplerule.pp#L57) is a subset of [the one in `nftables::rule`](https://github.com/duritong/puppet-nftables/blob/master/manifests/rule.pp#L6) (the first part of the big regexp which represents the chain is injected by `simplerule` from the value of a parameter). I don't find a way to reuse a regex declared as another type in a `Pattern` type so we'd have anyway to have to different types to maintain in this case.

In pseudocode something like:

```
type Nftables::LogicalRuleName = Pattern[/^[a-zA-Z0-9_]+(-\d+)?$/]
type Nftables::RuleName = Pattern[/^[a-zA-Z0-9_]+-#Nftables::LogicalRuleName_WithoutAnchors$/]
```

duritong: I usually start putting things into dedicated types, when either they become too long/complex or are used in multiple places.

duritong (approved):

manifests/simplerule.pp
bastelfreak: having it as a custom datatype also makes it easy to document it with puppet-strings and to test it. both are important for types with regular expressions, because they are hard to read

nbarrientos: I don't have a strong opinion. I'm happy to create two types (`Nftables::RuleName` and `Nftables::SimpleRuleName`, for instance). It won't help with having duplicate code but it might make reading and testing code easier, I guess. Would that be okay?

bastelfreak: @nbarrientos yes, please create those two

traylenator (approved):

PR #60, opened 2021-01-15
add some mail related outgoing rules

duritong:


traylenator:
Hi,
Can you add the new rules to `acceptance/all_rules_spec.rb` as well.

nbarrientos:
Also perhaps please also add the new classes to the reference, a `bundle exec rake strings:generate:reference` or similar should do.

manifests/rules/out/pop3.pp
```diff
@@ -0,0 +1,7 @@
+# manage out pop3
+class nftables::rules::out::pop3 {
+  nftables::rule {
+    'default_out-pop3':
+      content => 'tcp dport {145, 995} accept',
```
nbarrientos, line 5: Shouldn't this be `110` instead of `145`?

```
$ grep -i pop3 /etc/services 
pop3		110/tcp		pop-3		# POP version 3
pop3s		995/tcp				# POP-3 over SSL
````

Comment (duritong) [reply manifests/rules/out/pop3.pp:5]: Doh, good catch! I had mixed it up with imap :/

**Review by nbarrientos — APPROVED** [review]

File: manifests/rules/out/imap.pp [hunk 2]
```diff
@@ -0,0 +1,7 @@
+# allow outgoing imap
```
bastelfreak, line 1: please add puppet-strings style documentation for the new classes

duritong: Prefixing with `@summary` would be enough I guess?

bastelfreak: yes that should do it. afterwards run `bundle exec rake reference`

duritong: ok, it didn't make a difference

PR #62, opened 2021-01-18
Added Samba in rules

glpatcern:
These are the nft in rules to operate a Samba cluster, including `ctdb`. Tested at CERN.
The `ctdb` opening is optional and a parameter is provided for that.

N/A

### DIFF
--- a/manifests/rules/samba.pp
+++ b/manifests/rules/samba.pp
@@ -1,5 +1,7 @@
 # manage in samba/ctdb
-class nftables::rules::samba {
+class nftables::rules::samba (
+  Boolean $ctdb = false,
+) {
   nftables::rule {
     'default_in-netbios_tcp':
       content => 'tcp dport {139,445} accept',
@@ -10,8 +12,10 @@ class nftables::rules::samba {
       content => 'udp dport {137,138} accept',
   }
 
-  nftables::rule {
-    'default_in-ctdb':
-      content => 'tcp dport 4379 accept',
+  if ($ctdb) {
+    nftables::rule {
+      'default_in-ctdb':
+        content => 'tcp dport 4379 accept',
+    }
   }
 }
--- a/manifests/rules/samba.pp
+++ b/manifests/rules/samba.pp
@@ -13,7 +13,7 @@ class nftables::rules::samba (
       content => 'udp dport {137,138} accept',
   }
 
-  if ($ctdb) {
+  if $ctdb {
     nftables::rule {
       'default_in-ctdb':
         content => 'tcp dport 4379 accept',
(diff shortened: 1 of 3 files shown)


manifests/rules/samba.pp
```diff
@@ -0,0 +1,17 @@
...
+      content => 'tcp dport {139,445} accept',
+  }
+
+  nftables::rule {
+    'default_in-netbios_udp':
+      content => 'udp dport {137,138} accept',
+  }
+
+  nftables::rule {
+    'default_in-ctdb':
```
nbarrientos, line 14: If CTDB is not strictly necessary to run a Samba server I'd create a separate rule class for it (`nftables::rules::ctdb`) and move the responsibility of including it or not to the user of the module.

glpatcern: I was rather thinking to have this parameterized with a `boolean` (working on that, that's why I switched the PR to draft).

IMHO A ctdb rule alone makes little sense as you can't run it without samba.

manifests/rules/samba.pp
```diff
@@ -0,0 +1,23 @@
+# @summary manage Samba, the suite to allow Windows file sharing on Linux resources.
+# @params ctdb Enable ctdb-driven clustered Samba setups.
```
nbarrientos, line 2:
```suggestion
# @param ctdb Enable ctdb-driven clustered Samba setups.
```

nbarrientos (changes requested):

manifests/rules/samba.pp
```diff
@@ -0,0 +1,23 @@
+# @summary manage Samba, the suite to allow Windows file sharing on Linux resources.
+# @param ctdb Enable ctdb-driven clustered Samba setups.
+
```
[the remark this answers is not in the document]
glpatcern: oh gosh :-P ok, just squashed those last bits

nbarrientos (approved):
LGTM, thanks a lot for the patch! The `REFERENCE` can be regenerated later on. Asking for a second opinion, though.

manifests/rules/samba.pp
```diff
@@ -0,0 +1,22 @@
...
+    'default_in-netbios_tcp':
+      content => 'tcp dport {139,445} accept',
+  }
+
+  nftables::rule {
+    'default_in-netbios_udp':
+      content => 'udp dport {137,138} accept',
+  }
+
+  if ($ctdb) {
```
duritong, line 16: imho there is not need for the parantheses

duritong (changes requested):

manifests/rules/samba.pp
glpatcern: indeed, that was a left-over - removed now

PR #63, opened 2021-01-18
Add badges to README

traylenator:
Add badges to README.


traylenator:
The more interesting badge would be the build one but;

https://github.com/voxpupuli/puppet-nftables/workflows/.github/workflows/ci.yml/badge.svg

documented here 

https://docs.github.com/en/actions/managing-workflow-runs/adding-a-workflow-status-badge

does not want to work.

nbarrientos (approved):

PR #66, opened 2021-01-18
WIP: Add support for Debian

### LINKED ISSUE #65: Add support for Debian

- Puppet: 5
- Ruby: 2.5
- Distribution: Debian Buster
- Module version: 1.0.0

## Description

There are a few changes to make the module work on Debian

 * the nftables configuration file should /etc/nftables.conf instead of /etc/sysconfig/nftables.conf
 * the directory /etc/nftables is not created by the Debian package for nftables (managing it in the module would work?)
 * firewalld is not installed by default, so should be `false` instead of `mask` on Debian systems by default
 * (optional) for Debian Stretch, the version of facter shipped w/ the OS does not use structured facts. If it is desired to extend support to Debian Stretch, that case should be handled as well.
 
I have a WIP branch dev for which I will open a PR.

thanks

kienanstewart:
One thing I haven't figured out yet is how to fix the tests for Debian. I have added the defaults for the OS in data/os/Debian.yaml, but they do not seem to be used.

Adds support for Debian Buster (10).

Fixes #65

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -153,13 +153,13 @@ class nftables (
     restart    => '/usr/bin/systemctl reload nftables',
   }
 
+  $puppet_nft_vars = {
+    'nftables_configuration_path' => $nftables_configuration_path,
+  }
   systemd::dropin_file { 'puppet_nft.conf':
     ensure  => present,
     unit    => 'nftables.service',
-    content => file(
-      "nftables/systemd/puppet_nft.${facts['os']['family']}.conf",
-      'nftables/systemd/puppet_nft.conf'
-    ),
+    content => epp('nftables/systemd/puppet_nft.conf.epp', $puppet_nft_vars),
     notify  => Service['nftables'],
   }
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -80,6 +80,9 @@
 # @param rules
 #   Specify hashes of `nftables::rule`s via hiera
 #
+# @param nftables_configuration_path
+#   The absolute path to the principal nftables configuration file.
+#
 class nftables (
   Boolean $in_ssh = true,
   Boolean $in_icmp = true,
@@ -99,19 +102,23 @@ class nftables (
   Variant[Boolean[false], Pattern[/icmp(v6|x)? type .+|tcp reset/]] $reject_with = 'icmpx type port-unreachable',
   Variant[Boolean[false], Enum['mask']] $firewalld_enable = 'mask',
   Optional[Array[Pattern[/^(ip|ip6|inet)-[-a-zA-Z0-9_]+$/],1]] $noflush_tables = undef,
+  Stdlib::AbsolutePath $nftables_configuration_path = '/etc/sysconfig/nftables.conf',
 ) {
   package { 'nftables':
     ensure => installed,
   } -> file_line {
     'enable_nftables':
       line   => 'include "/etc/nftables/puppet.nft"',
-      path   => '/etc/sysconfig/nftables.conf',
+      path   => $nftables_configuration_path,
       notify => Service['nftables'],
   } -> file {
     default:
       owner => 'root',
       group => 'root',
       mode  => '0640';
+    '/etc/nftables':
+      ensure => directory,
+      mode   => '0750';
     '/etc/nftables/puppet-preflight':
       ensure  => directory,
       mode    => '0750',
@@ -149,7 +156,10 @@ class nftables (
   systemd::dropin_file { 'puppet_nft.conf':
     ensure  => present,
     unit    => 'nftables.service',
-    content => file('nftables/systemd/puppet_nft.conf'),
+    content => file(
+      "nftables/systemd/puppet_nft.${facts['os']['family']}.conf",
+      'nftables/systemd/puppet_nft.conf'
+    ),
     notify  => Service['nftables'],
   }
 
--- a/metadata.json
+++ b/metadata.json
@@ -45,6 +45,12 @@
       "operatingsystemrelease": [
         "8"
       ]
+    },
+    {
+      "operatingsystem": "Debian",
+      "operatingsystemrelease": [
+        "10"
+      ]
     }
   ],
   "requirements": [
--- a/spec/classes/nftables_spec.rb
+++ b/spec/classes/nftables_spec.rb
@@ -12,6 +12,15 @@ describe 'nftables' do
       it { is_expected.to contain_package('nftables') }
 
       it {
+        is_expected.to contain_file('/etc/nftables').with(
+          ensure: 'directory',
+          owner:  'root',
+          group:  'root',
+          mode:    '0750',
+        )
+      }
+
+      it {
         is_expected.to contain_file('/etc/nftables/puppet.nft').with(
           ensure: 'file',
           owner:  'root',
@@ -71,18 +80,34 @@ describe 'nftables' do
         )
       }
 
-      it {
-        is_expected.to contain_systemd__dropin_file('puppet_nft.conf').with(
-          content: %r{^ExecReload=/sbin/nft -I /etc/nftables/puppet -f /etc/sysconfig/nftables.conf$},
-        )
-      }
+      if os_facts[:os]['family']  == 'Debian'
+        it {
+          is_expected.to contain_systemd__dropin_file('puppet_nft.conf').with(
+            content: %r{^ExecReload=/sbin/nft -I /etc/nftables/puppet -f /etc/nftables.conf$},
+          )
+        }
 
-      it {
-        is_expected.to contain_service('firewalld').with(
-          ensure: 'stopped',
-          enable: 'mask',
-        )
-      }
+        it {
+          is_expected.to contain_service('firewalld').with(
+            ensure: 'stopped',
+            enable: false,
+          )
+        }
+      else
+        it {
+          is_expected.to contain_systemd__dropin_file('puppet_nft.conf').with(
+            content: %r{^ExecReload=/sbin/nft -I /etc/nftables/puppet -f /etc/sysconfig/nftables.conf$},
+          )
+        }
+
+        it {
+          is_expected.to contain_service('firewalld').with(
+            ensure: 'stopped',
+            enable: 'mask',
+          )
+        }
+      end
+
       it { is_expected.to contain_class('nftables::rules::out::http') }
       it { is_expected.to contain_class('nftables::rules::out::https') }
       it { is_expected.to contain_class('nftables::rules::out::dns') }
--- a/data/os/Debian.yaml
+++ b/data/os/Debian.yaml
@@ -1,0 +1,6 @@
+---
+
+# firewalld is not installed by default in Debian
+firewalld_enable: false
+
+nftables_configuration_path: /etc/nftables.conf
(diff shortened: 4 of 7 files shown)


manifests/init.pp
```diff
@@ -99,19 +102,23 @@
...
+      path   => $nftables_configuration_path,
       notify => Service['nftables'],
   } -> file {
     default:
       owner => 'root',
       group => 'root',
       mode  => '0640';
+    '/etc/nftables':
+      ensure => directory,
+      mode   => '0750';
```
duritong, line 121: should we purge it? maybe only on debian?

spec/classes/nftables_spec.rb
```diff
@@ -71,18 +80,34 @@
         )
       }
 
-      it {
-        is_expected.to contain_systemd__dropin_file('puppet_nft.conf').with(
-          content: %r{^ExecReload=/sbin/nft -I /etc/nftables/puppet -f /etc/sysconfig/nftables.conf$},
-        )
-      }
+      if os_facts[:os]['family']  == 'Debian'
```
duritong, line 83: once access as through a symbol and once as string? Is this really needed like that?

kienanstewart: If `:os` is a string `'os'`, the code fails with ```
Failure/Error: if os_facts['os']['family']  == 'Debian'

NoMethodError:
  undefined method `[]' for nil:NilClass
``` 

However, it seems both than be symbols.

File: manifests/init.pp (hunk 1 above)
Comment (kienanstewart) [reply manifests/init.pp:121]: It seems reasonable to me that it be purged, although I do not know what gets put in the by other OSes.

File: spec/classes/nftables_spec.rb (hunk 2 above)
Comment (kienanstewart) [reply spec/classes/nftables_spec.rb]: I've updated it to use symbols for both parts of the lookup

File: manifests/init.pp (hunk 1 above)
Comment (kienanstewart) [reply manifests/init.pp:121]: Upon some further reflexion, I'm not sure it is that useful to purge the directory since the nftables service is specifically using `/etc/nftables/puppet.nft`

Comment (duritong) [reply manifests/init.pp:121]: I guess, it's probably fine to not do. Just an idea that I wanted to throw in.

File: manifests/init.pp [hunk 3]
```diff
@@ -99,19 +102,23 @@
   Variant[Boolean[false], Pattern[/icmp(v6|x)? type .+|tcp reset/]] $reject_with = 'icmpx type port-unreachable',
   Variant[Boolean[false], Enum['mask']] $firewalld_enable = 'mask',
   Optional[Array[Pattern[/^(ip|ip6|inet)-[-a-zA-Z0-9_]+$/],1]] $noflush_tables = undef,
+  Stdlib::AbsolutePath $nftables_configuration_path = '/etc/sysconfig/nftables.conf',
```
duritong, line 105: Unixpath is probably fine, I don't think there is nftables on Windows?

spec/spec_helper.rb
```diff
@@ -20,3 +20,9 @@
     end
   end
 end
+
+# This loads the module data, needed to test different OSes.
+#RSpec.shared_context "hiera", :shared_context => :metadata do
+#  let (:hiera_config) { 'hiera.yaml' }
+#  Hiera.new({:config => 'hiera.yaml'})
+#end
```
duritong, line 28: can you remove that change. it would anyway be overwritten by modulesync.

duritong (changes requested):

manifests/init.pp
traylenator, line 105: I think just `configugration_path` is enough as a variable name.

spec/spec_helper.rb
traylenator: I don't believe it is needed for module level data ? If it is there is a 

```
<%- if @configs['hiera_config'] -%>

RSpec.configure do |c|
  c.hiera_config = <%= @configs['hiera_config'] %> 
end
<%- end -%>
```
in the template for `spec_helper.rb`

traylenator (commented):
Nice to have Debian support.

spec/spec_helper.rb
kienanstewart: I had added that becuase I missed the fact that I didn't name the keys in Debian.yaml properly, which I fixed in https://github.com/voxpupuli/puppet-nftables/pull/66/commits/f8c4097f31bc18500a1c7e05dcd203a3e3b5207b

I will remove the commented code.

manifests/init.pp
kienanstewart: It is linux specific, I will change it to Unixpath

kienanstewart: Sounds good, I've made that change

manifests/init.pp
```diff
@@ -99,19 +102,23 @@
   Variant[Boolean[false], Pattern[/icmp(v6|x)? type .+|tcp reset/]] $reject_with = 'icmpx type port-unreachable',
   Variant[Boolean[false], Enum['mask']] $firewalld_enable = 'mask',
   Optional[Array[Pattern[/^(ip|ip6|inet)-[-a-zA-Z0-9_]+$/],1]] $noflush_tables = undef,
+  Stdlib::Unixpath $configuration_path = '/etc/sysconfig/nftables.conf',
```
bastelfreak, line 105: can you move this to hiera as well? If we generate the docs with puppet-strings, the data from the init.pp will land in the REFERENCE.md. People will then assume that `/etc/sysconfig/nftables.conf` will be used on all systems.

kienanstewart: For sure

spec/classes/nftables_spec.rb
traylenator, line 83:
```suggestion
      if os_facts[:os]['family'] == 'Debian'
```

Should fix rubocop

traylenator: `/usr/bin/echo` -> `/bin/echo` in `init.pp`

nbarrientos:
The syntax used in https://github.com/voxpupuli/puppet-nftables/blob/master/manifests/rules/nfs3.pp#L5 (more specifically the transport header keyword), does not seem to be recognised by the version of nftables shipped by Debian 10:

https://github.com/voxpupuli/puppet-nftables/pull/66/checks?check_run_id=1743370200#step:5:514

`0.9.3`' s manpage (shipped by CentOS8) reads:

```
       Matching destination port of both UDP and TCP.

           inet filter input meta l4proto {tcp, udp} @th,16,16 { 53, 80 }

       The above can also be written as

           inet filter input meta l4proto {tcp, udp} th dport { 53, 80 }
```

whereas that does not seem to be mentioned in `0.9.0`'s (Debian 10.7).

nbarrientos:
I think these four need tweaking, too:

https://github.com/voxpupuli/puppet-nftables/blob/master/spec/acceptance/all_rules_spec.rb#L78
https://github.com/voxpupuli/puppet-nftables/blob/master/spec/acceptance/all_rules_spec.rb#L80
https://github.com/voxpupuli/puppet-nftables/blob/master/spec/acceptance/default_spec.rb#L19
https://github.com/voxpupuli/puppet-nftables/blob/master/spec/acceptance/default_spec.rb#L21

as if I understand the proposed patch correctly `/etc/nftables.conf` is used as the root configuration file instead of `/etc/sysconfig/nftables.conf`.

kienanstewart:
> The syntax used in https://github.com/voxpupuli/puppet-nftables/blob/master/manifests/rules/nfs3.pp#L5 (more specifically the transport header keyword), does not seem to be recognised by the version of nftables shipped by Debian 10:
> 
> https://github.com/voxpupuli/puppet-nftables/pull/66/checks?check_run_id=1743370200#step:5:514
> 
> `0.9.3`' s manpage (shipped by CentOS8) reads:
> 
> ```
>        Matching destination port of both UDP and TCP.
> 
>            inet filter input meta l4proto {tcp, udp} @th,16,16 { 53, 80 }
> 
>        The above can also be written as
> 
>            inet filter input meta l4proto {tcp, udp} th dport { 53, 80 }
> ```
> 
> whereas that does not seem to be mentioned in `0.9.0`'s (Debian 10.7).

It looks like `nfs` is not recognized either. Something like `@th,16,16 2049` instead of `th dport nfs` works. I am not sure if `nfs` always evaluates to 2049, or how that lookup happens.

kienanstewart:
I don't see how to easily change the acceptance tests yet. Do you have an idea off-hand? In any case I'll keep at it

duritong:
> It looks like `nfs` is not recognized either. Something like `@th,16,16 2049` instead of `th dport nfs` works. I am not sure if `nfs` always evaluates to 2049, or how that lookup happens.

Named ports nfs are mapped by their entries in `/etc/services` are they really different on debian than on EL based systems? Wow...

nbarrientos:
> It looks like nfs is not recognized either. Something like @th,16,16 2049 instead of th dport nfs works. I am not sure if nfs always evaluates to 2049, or how that lookup happens.

/cc @dvanders

duritong:
What is left here?

kienanstewart:
There's still some fixes for the the CI tests in live systems. I don't really know how to tackle changing them, but that's more a result of not being able to reproduce the testing environment locally and just requires more of my time.

A larger issue I think is how to approach the nfs rulesets which aren't compatible. One way would simply to be case against the OS version, but it looks like it's more a question of the nftables version - 0.9.0 not supporting the same syntax as 0.9.3. I think the commit I added (https://github.com/voxpupuli/puppet-nftables/pull/66/commits/df9a09bb3b8bde7c47097fee9779a12ee2f1bfd3) is flawed since the nfs port may vary from system to system.

traylenator:
We could extend the existing nftables fact to contain a `nftables::version`

traylenator:
One idea here would be to switch the metadata to support  ubuntu 20.04 for instance which has

```
# nft --version
nftables v0.9.3 (Topsy)
```

And then the narrower problem of older nft can be solved outside this ticket.

traylenator:
> We could extend the existing nftables fact to contain a `nftables::version`

https://github.com/voxpupuli/puppet-nftables/pull/84  adds `nftables.version`

rvstaveren:
Just started to roll out puppet (again) and taking Debian into account and was happy to see there is already an issue running for this.

roock:
@ kienanstewart are you still working/interested  in this?
I continued your work, updated it to latest master and fixed the broken CI tests (see https://github.com/roock/puppet-nftables/pull/1). Do you wanna integrate my changes so we get that into master or should I create a new PR?

manifests/rules/nfs3.pp
```diff
@@ -2,6 +2,7 @@
 class nftables::rules::nfs3 {
   nftables::rule {
     'default_in-nfs3':
-      content => 'meta l4proto { tcp, udp } th dport nfs accept comment "Accept NFS3"',
+      # To support nftables >= 0.9.0, but < 0.9.3, `@th,16,16 2049` is used instead of `th dport nfs`
```
hashworks, line 5: Debian Stable (11) is using 0.9.8, so this should be adjusted.

metadata.json
```diff
@@ -45,6 +45,12 @@
       "operatingsystemrelease": [
         "8"
       ]
+    },
+    {
+      "operatingsystem": "Debian",
+      "operatingsystemrelease": [
+        "10"
```
hashworks, line 52: We should support the latest stable version (11)

hashworks (changes requested):

hashworks:
Since #124 has been merged this needs to be rebased.

kienanstewart:
@bastelfreak thanks! and sorry I wasn't able to get it into a working state

PR #68, opened 2021-01-26
Improve nftables::rule's documentation

nbarrientos:
This patchset documents a bit better the parameters of the defined type and adds some examples.

### DIFF
--- a/manifests/rule.pp
+++ b/manifests/rule.pp
@@ -1,15 +1,44 @@
-# manage a chain rule
-# Name should be:
-#   CHAIN_NAME-rulename
+# @summary Provides an interface to create a firewall rule
+#
+# @example add a rule named 'myhttp' to the 'default_in' chain to allow incoming traffic to TCP port 80
+#  nftables::rule {
+#    'default_in-myhttp':
+#      content => 'tcp dport 80 accept',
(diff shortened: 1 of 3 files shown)


manifests/rule.pp
```diff
@@ -1,15 +1,44 @@
...
+#      content => 'counter',
+#      table   => 'ip6-nat'
+#  }
+#
+# @param ensure
+#   Should the rule be created.
+#
+# @param rulename
+#   The symbolic name for the rule and to what chain to add it. The
+#   format is defined by the Nftables::RuleName type.
```
traylenator, line 21: Last sentence seems unnessary as it displayed as the type anyway.

nbarrientos: Yup, kinda on purpose. I had the newbies in mind when I wrote that as the generated reference does not automatically include a link to the documentation of the type in that particular point so perhaps adding that extra bit might help making the reader think that they should read the documentation of the type for more information about the syntax. It can go, no problem.

duritong: it's ok for me to keep it.

duritong (approved):

traylenator (approved):

PR #73, opened 2021-02-11
start declaring the 'global' chain with module resources

lelutin:
the 'global' chain is a vestigial piece of early development on this
module, but it can be useful for creating fast short-circuits like
blocking traffic that match a certain set of IPs.

in the current state we can't inject rules inside the 'global' chain
since it's unknown to puppet. so let's remove the hard-coded definition
and use a puppet resource to declare it.

### DIFF
--- a/files/config/puppet-inet-filter.nft
+++ b/files/config/puppet-inet-filter.nft
@@ -1,5 +1,1 @@
   include "inet-filter-chain-*.nft"
-
-  # something we want for all
-  chain global {
-  }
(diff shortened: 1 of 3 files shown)


nbarrientos:
Sound good. Could perhaps some extra test coverage be added to `spec/classes/inet_filter_spec.rb` at the same time to make sure that the chain is created and cover regressions?

duritong:
yeah, let's add a spec and merge.

lelutin:
@nbarrientos ah yes good point, I forgot to add tests. I've just sent the added tests, let's see how CI likes them (at least locally they were failing on the master branch and successful on top of this branch)

nbarrientos (approved):
LGTM

duritong (approved):

duritong:
thank you!

PR #75, opened 2021-03-01
fix #74 - ensure table are initialized before flushing them

### LINKED ISSUE #74: nftables service is broken after reboot
I am having the issue that after rebooting a node the nftables service is broken:

```
# systemctl status nftables.service 
● nftables.service - Netfilter Tables
   Loaded: loaded (/usr/lib/systemd/system/nftables.service; enabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/nftables.service.d
           └─puppet_nft.conf
   Active: failed (Result: exit-code) since Mon 2021-03-01 14:21:53 UTC; 11min ago
     Docs: man:nft(8)
  Process: 759 ExecStart=/sbin/nft -I /etc/nftables/puppet -f /etc/sysconfig/nftables.conf (code=exited, status=1/FAILURE)
 Main PID: 759 (code=exited, status=1/FAILURE)

Mar 01 14:21:53 example.glei.ch nft[759]: /etc/nftables/puppet.nft:8:26-44: Error: Could not process rule: No such file or directory
Mar 01 14:21:53 example.glei.ch nft[759]: flush table inet filter; flush table ip nat; flush table ip6 nat
Mar 01 14:21:53 example.glei.ch nft[759]:                          ^^^^^^^^^^^^^^^^^^^
Mar 01 14:21:53 example.glei.ch nft[759]: In file included from /etc/sysconfig/nftables.conf:4:1-35:
Mar 01 14:21:53 example.glei.ch nft[759]: /etc/nftables/puppet.nft:8:46-65: Error: Could not process rule: No such file or directory
Mar 01 14:21:53 example.glei.ch nft[759]: flush table inet filter; flush table ip nat; flush table ip6 nat
Mar 01 14:21:53 example.glei.ch nft[759]:                                              ^^^^^^^^^^^^^^^^^^^^
Mar 01 14:21:53 example.glei.ch systemd[1]: nftables.service: Main process exited, code=exited, status=1/FAILURE
Mar 01 14:21:53 example.glei.ch systemd[1]: nftables.service: Failed with result 'exit-code'.
Mar 01 14:21:53 example.glei.ch systemd[1]: Failed to start Netfilter Tables.
```

It will go back in a working by running puppet twice (!).

First it will remove the flush table cmds (since the fact returns an empty list of tables) and thus the nft rule set become applicable again.
On the second run it will re-add the tables.

We should not try to flush the tables when they are not present!

duritong:
fix #74

### DIFF
--- a/spec/classes/nftables_spec.rb
+++ b/spec/classes/nftables_spec.rb
@@ -192,6 +192,10 @@ describe 'nftables' do
 
           it {
             is_expected.to contain_file('/etc/nftables/puppet-preflight.nft').
+              with_content(%r{^table inet abc \{\}$})
+          }
+          it {
+            is_expected.to contain_file('/etc/nftables/puppet-preflight.nft').
               with_content(%r{^flush table inet abc$})
           }
         end
@@ -202,7 +206,19 @@ describe 'nftables' do
 
           it {
             is_expected.to contain_file('/etc/nftables/puppet-preflight.nft').
-              with_content(%r{^flush table inet abc; flush table inet ijk$})
+              with_content(%r{^table inet abc \{\}$})
+          }
+          it {
+            is_expected.to contain_file('/etc/nftables/puppet-preflight.nft').
+              with_content(%r{^flush table inet abc$})
+          }
+          it {
+            is_expected.to contain_file('/etc/nftables/puppet-preflight.nft').
+              with_content(%r{^table inet ijk \{\}$})
+          }
+          it {
+            is_expected.to contain_file('/etc/nftables/puppet-preflight.nft').
+              with_content(%r{^flush table inet ijk$})
           }
         end
       end
--- a/templates/config/puppet.nft.epp
+++ b/templates/config/puppet.nft.epp
@@ -5,8 +5,8 @@
 <%-
 if $noflush and $facts['nftables'] and $facts['nftables']['tables'] {
  $_flush_command = $facts['nftables']['tables'].filter |$_tab| { ! ($_tab in $noflush) }.map |$_table| {
-   "flush table ${regsubst($_table,'-',' ')}"
- }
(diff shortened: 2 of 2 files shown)


nbarrientos:
Hehe, hacky fix. You're lucky that executing `table inet foo {}` when the table `foo` exists does not make `nftables` complain 😆

nbarrientos (approved):

duritong:
I didn't really see a way to conditionally check for the existence of the table

traylenator:
I noticed this myself after adding that original patch - but could not think of a good way.
The other option is just to completly remove the flush from start up since it is redundant. We would rely on the flush in reload only.

Lets go with this unless you think removing the flush completly is a better idea.

traylenator (approved):

nbarrientos:
Perhaps a good time to create a release, there are two interesting fixes in master now.

duritong:
Question: do you know why the author was rewritten when merged? @traylenator

traylenator:
This commt  https://github.com/voxpupuli/puppet-nftables/commit/92e0fcb6d4afb8d771e401cf050dc5c4da836f48 is written by you ?

duritong:
I pushed & proposed to merge a4c3c76b2b221e3c3b53d906213c8f98de3dc152 but merged was 92e0fcb6d4afb8d771e401cf050dc5c4da836f48 (also on master is this) - which is content-wise the same commit, except that the author got re-written, thus a different commit-sha.

So I was wondering why the author got re-written and whether you clicked anything special.

Background: When I now try to merge back upstream (this repo) I need either to merge or throw away my old commit, but this means I need to internally force push, which is also weird. Otherwise github started doing that which would be a really weird thing to start doing...

traylenator:
> whether you clicked anything special.

I clicked what ever was the default option when doing the merge assuming GH knows best. 
I will take note next time what the options are and try and preserve.

PR #80, opened 2021-03-08
Add Docker-CE default rules

luisfdez:
#### Add Docker-CE default rules

This Pull Request adds the set of rules and chains that docker-ce daemon sets on a CentOS8 machine by default.

These rules are defined in `ip-filter`, not playing nicely with the defaults shipped for `inet-filter`.

This merge request adds the set of rules that Docker deploys by default. I don't know how relevant all of them are, but I just followed what the daemon was doing.

None

### DIFF
--- a/manifests/rules/docker_ce.pp
+++ b/manifests/rules/docker_ce.pp
@@ -9,9 +9,9 @@
 # When using this class 'docker::iptables: false' should be set.
 #
 # @param docker_interface
-#   Interface name used by docker. It defaults to docker0.
+#   Interface name used by docker.
 # @param docker_prefix
-#   The address space used by docker. It defaults to 172.17.0.0/16.
+#   The address space used by docker.
 #
 class nftables::rules::docker_ce (
   String[1]                     $docker_interface = 'docker0',
--- a/manifests/rules/docker_ce.pp
+++ b/manifests/rules/docker_ce.pp
@@ -91,9 +91,9 @@ class nftables::rules::docker_ce (
       table   => 'ip-nat',
       content => 'fib daddr type local counter jump DOCKER';
     'OUTPUT-jump_docker@ip-nat':
-      rule_name => 'OUTPUT-jump_docker',
-      table     => 'ip-nat',
-      content   => 'ip daddr != 127.0.0.0/8 fib daddr type local counter jump DOCKER';
+      rulename => 'OUTPUT-jump_docker',
+      table    => 'ip-nat',
+      content  => 'ip daddr != 127.0.0.0/8 fib daddr type local counter jump DOCKER';
     'DOCKER-counter':
       table   => 'ip-nat',
       content => "iifname \"${docker_interface}\" counter return";
--- a/manifests/rules/docker_ce.pp
+++ b/manifests/rules/docker_ce.pp
@@ -1,0 +1,111 @@
+# @summary Default firewall configuration for Docker-CE
+#
+# The configuration distributed in this class represents the default firewall
+# configuration done by docker-ce when the iptables integration is enabled.
+#
+# This class is needed as the default docker-ce rules added to ip-filter conflict
+# with the inet-filter forward rules set by default in this module.
+#
+# When using this class 'docker::iptables: false' should be set.
+#
+# @param docker_interface
+#   Interface name used by docker. It defaults to docker0.
+# @param docker_prefix
+#   The address space used by docker. It defaults to 172.17.0.0/16.
+#
+class nftables::rules::docker_ce (
+  String[1]                     $docker_interface = 'docker0',
+  Stdlib::IP::Address::V4::CIDR $docker_prefix    = '172.17.0.0/16',
+) {
+  #
+  # inet-filter
+  #
+
+  nftables::chain {
+    'DOCKER': ;
+    'DOCKER_ISOLATION_STAGE_1': ;
+    'DOCKER_ISOLATION_STAGE_2': ;
+    'DOCKER_USER': ;
+  }
+
+  nftables::rule {
+    'DOCKER_ISOLATION_STAGE_1-iifname':
+      order   => '01',
+      content => "iifname \"${docker_interface}\" oifname != \"${docker_interface}\" counter jump DOCKER_ISOLATION_STAGE_2";
+    'DOCKER_ISOLATION_STAGE_1-counter':
+      order   => '02',
+      content => 'counter return';
+    'DOCKER_ISOLATION_STAGE_2-drop':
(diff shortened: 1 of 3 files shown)


nbarrientos:
Thanks!

Could you please amend the lint errors the CI suggests and also add the new class to the [`all_rules` acceptance test](https://github.com/voxpupuli/puppet-nftables/blob/master/spec/acceptance/all_rules_spec.rb)?

luisfdez:
Thanks Nacho, I have fixed the lint errors and added the new class to the acceptance tests.

nbarrientos:
Thanks! Maybe it'd be a good idea to add some unit test coverage to the new class to help us catching regressions in the future and to make sure for instance that the parameters are correctly taken into account. I've written some that you can just import here if you like them or use as skeleton for your own:

https://github.com/cernops/puppet-nftables/commit/2c53a270ec92efc454c6824aa94eebfe7d6183f0

manifests/rules/docker_ce.pp
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  nftables::rule {
+    'POSTROUTING-docker':
+      table   => 'ip-nat',
+      content => "oifname != \"${docker_interface}\" ip saddr ${docker_prefix} counter masquerade";
+    'PREROUTING-docker':
+      table   => 'ip-nat',
+      content => 'fib daddr type local counter jump DOCKER';
+    'OUTPUT-jump_docker':
```
nbarrientos, line 93:
```suggestion
    'OUTPUT-jump_docker@ip-nat':
```
There's an `OUTPUT` chain in `inet-filter`.

manifests/rules/docker_ce.pp
```diff
@@ -0,0 +1,110 @@
...
+#
+# When using this class 'docker::iptables: false' should be set.
+#
+# @param docker_interface
+#   Interface name used by docker. It defaults to docker0.
+# @param docker_prefix
+#   The address space used by docker. It defaults to 172.17.0.0/16.
+#
+class nftables::rules::docker_ce (
+  String                        $docker_interface = 'docker0',
```
traylenator, line 17:
```suggestion
  String[1]                        $docker_interface = 'docker0',
```
string needs to have length.

traylenator (changes requested):

manifests/rules/docker_ce.pp
luisfdez: done.

manifests/rules/docker_ce.pp
luisfdez: Thanks, just changed that. It should work both ways, but it does help with readability.

luisfdez:
> Thanks! Maybe it'd be a good idea to add some unit test coverage to the new class to help us catching regressions in the future and to make sure for instance that the parameters are correctly taken into account. I've written some that you can just import here if you like them or use as skeleton for your own:
> 
> [cernops@2b7935f](https://github.com/cernops/puppet-nftables/commit/2b7935f17d9f84447644db88361e7e1a905eb54b)

Thanks Nacho, I've taken your example and expanded with other rules.

luisfdez:
> Presumably as machines are created and destroyed new rules will be created and deleted ?
> Do we need some docs to to set the `noflush_tables` parameter ?

I don't know what would be the right approach to make it rock solid. In principle in a default/basic usage there shouldn't be rules added as it relies on the interface and the address space. In any case, I'm sure that there will be cases not covered.

traylenator (dismissed):

tmanninger:
If the OUTPUT-NAT chain is already defined, i get the a duplication declaration error.
The declaration should be optional.

manifests/rules/docker_ce.pp
```diff
@@ -0,0 +1,111 @@
...
+# The configuration distributed in this class represents the default firewall
+# configuration done by docker-ce when the iptables integration is enabled.
+#
+# This class is needed as the default docker-ce rules added to ip-filter conflict
+# with the inet-filter forward rules set by default in this module.
+#
+# When using this class 'docker::iptables: false' should be set.
+#
+# @param docker_interface
+#   Interface name used by docker. It defaults to docker0.
```
nbarrientos, line 12:
```suggestion
#   Interface name used by docker.
```

The default value will be automatically picked up from the code when generating `REFERENCE.md` so there's no need to specify it here again.

manifests/rules/docker_ce.pp
```diff
@@ -0,0 +1,111 @@
...
+# @param docker_prefix
+#   The address space used by docker. It defaults to 172.17.0.0/16.
```
nbarrientos, line 14:
```suggestion
#   The address space used by docker.
```
The default value will be automatically picked up from the code when generating `REFERENCE.md` so there's no need to specify it here again.

nbarrientos (changes requested):

manifests/rules/docker_ce.pp
luisfdez: Done.

manifests/rules/docker_ce.pp
luisfdez: Done.

manifests/rules/docker_ce.pp
```diff
@@ -0,0 +1,124 @@
...
+      'DOCKER_ISOLATION_STAGE_1': ;
+      'DOCKER_ISOLATION_STAGE_2': ;
+      'DOCKER_USER': ;
+    }
+  }
+
+  nftables::rule {
+    'DOCKER_ISOLATION_STAGE_1-iifname':
+      order   => '01',
+      content => "iifname \"${docker_interface}\" oifname != \"${docker_interface}\" counter jump DOCKER_ISOLATION_STAGE_2";
```
bastelfreak, line 40: same question as in the other PR: are the escaped quotes required / can we use single quotes?

luisfdez:
> If the OUTPUT-NAT chain is already defined, i get the a duplication declaration error.
> The declaration should be optional.

@tmanninger I have added a couple of class parameters to handle the creation of chains. Does it look good to you now? Thanks for taking a look :-)

manifests/rules/docker_ce.pp
luisfdez: Normal interface names: eth0, docker0 could go without double quotes. However, if for some reason some uncommon interface name is used double quotes are needed. This approach should be safer for those uncommon cases.
There is an example of this in `man nft` describing the string type.

bastelfreak (approved):

nbarrientos (approved):

duritong (approved):

PR #104, opened 2021-07-24
fix datatype for $table and $dport

bastelfreak:


bastelfreak:
required for https://github.com/voxpupuli/puppet-nftables/runs/3150689539?check_suite_focus=true to pass

smortex (approved):
Makes sense!

PR #105, opened 2021-07-26
support a different table name for 'nat'

figless:
- Some applications (such as libvirt) still use iptables to inject firewall
  rules
- iptables will refuse to update tables that were initially created with nft
- This commit allows defining the name of the 'nat' table in order to avoid
  namespace conflicts


manifests/init.pp
```diff
@@ -99,6 +102,7 @@
   Hash $rules = {},
   Hash $sets = {},
   String $log_prefix = '[nftables] %<chain>s %<comment>s',
+  String $nat_table_name = 'nat',
```
traylenator, line 105:
```suggestion
  String[1] $nat_table_name = 'nat',
```

manifests/rules/qemu.pp
```diff
@@ -41,7 +41,7 @@
   Boolean                                 $forward_traffic   = true,
   Boolean                                 $internal_traffic  = true,
   Boolean                                 $masquerade        = true,
-) {
+) inherits nftables {
```
nbarrientos, line 44: It's safe to assume that `nftables` will be present when including classes in `nftables::rules::*`. For the test to pass, just add:

```ruby
let(:pre_condition) { 'include nftables' }
```

before defining the contexts, something like:

```ruby
describe 'nftables::rules::qemu' do
  on_supported_os.each do |os, os_facts|
    context "on #{os}" do
      let(:facts) { os_facts }

      let(:pre_condition) { 'include nftables' }

      context 'default options' do
        it { is_expected.to compile }
...
```
https://rspec-puppet.com/documentation/classes/#specifying-code-to-include-before

This suggestion applies to more rules that you have patched, please review them.

manifests/rules/docker_ce.pp
```diff
@@ -80,44 +80,44 @@
...
   nftables::rule {
     'POSTROUTING-docker':
-      table   => 'ip-nat',
+      table   => "ip-${nftables::nat_table_name}",
       content => "oifname != \"${docker_interface}\" ip saddr ${docker_prefix} counter masquerade";
     'PREROUTING-docker':
-      table   => 'ip-nat',
+      table   => "ip-${nftables::nat_table_name}",
       content => 'fib daddr type local counter jump DOCKER';
     'OUTPUT-jump_docker@ip-nat':
```
nbarrientos, line 106: I'd suggest to make the resource name match the table name using the variable (several cases in this file).

nbarrientos:
Would you mind adding a test case to `ip_nat_spec.rb` with `nat_table_table` set to a different value?

nbarrientos:
Would you mind adding a test case to `acceptance/default_spec.rb` with `nat_table_table` set to a different value to make sure that the include chains are correctly generated and they can be loaded? Similar to `with totally empty firewall` but setting the new variable, just to make sure that the service starts (the rules validate).

spec/classes/rules/qemu_spec.rb
```diff
@@ -77,7 +77,6 @@
         end
 
         it { is_expected.to compile }
-        it { is_expected.to have_resource_count(0) }
```
nbarrientos, line 80: If I'm not wrong you're removing this assert because `nftables` is present now when the class is tested. However, removing it reduces test coverage as the point of this assert was to make sure that `nftables::rules::qemu` didn't add any resource if all the toggles were off. I'd suggest to then actively assert that neither of the rules created by `nftables::rules::qemu` are present to have the same test coverage as before the patch.

PR #107, opened 2021-08-26
support a different table name for 'nat'

figless:
- Some applications (such as libvirt) still use iptables to inject firewall
  rules
- iptables will refuse to update tables that were initially created with nft
- This commit allows defining the name of the 'nat' table in order to avoid
  namespace conflicts

Note: pull request https://github.com/voxpupuli/puppet-nftables/pull/105 was raised previously, however was closed due to an erroneous force push to the forked project. Please check this pull request for the full change history.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -43,6 +43,9 @@
 # @param nat
 #   Add default tables and chains to process NAT traffic.
 #
+# @param nat_table_name
+#   The name of the 'nat' table.
+#
 # @param sets
 #   Allows sourcing set definitions directly from Hiera.
 #
@@ -99,6 +102,7 @@ class nftables (
   Hash $rules = {},
   Hash $sets = {},
   String $log_prefix = '[nftables] %<chain>s %<comment>s',
+  String[1] $nat_table_name = 'nat',
   Variant[Boolean[false], String] $log_limit = '3/minute burst 5 packets',
   Variant[Boolean[false], Pattern[/icmp(v6|x)? type .+|tcp reset/]] $reject_with = 'icmpx type port-unreachable',
   Variant[Boolean[false], Enum['mask']] $firewalld_enable = 'mask',
--- a/spec/acceptance/default_spec.rb
+++ b/spec/acceptance/default_spec.rb
@@ -111,4 +111,34 @@ describe 'nftables class' do
       it { is_expected.to be_enabled }
     end
   end
+  context 'with custom nat_table_name' do
+    it 'no rules validate okay' do
+      pp = <<-EOS
+      class{'nftables':
+        firewalld_enable => false,
+        nat => true,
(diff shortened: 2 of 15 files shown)


manifests/init.pp
```diff
@@ -43,6 +43,9 @@
 # @param nat
 #   Add default tables and chains to process NAT traffic.
 #
+# @param nat_table_name
+#   The name of the 'nat' table: defaults to 'nat'
```
nbarrientos, line 47: There's no need to describe here the default value, it will be automatically described in `REFERENCE.md` after regenerating the doc strings.

spec/acceptance/default_spec.rb
```diff
@@ -111,4 +111,34 @@ class { 'nftables':
       it { is_expected.to be_enabled }
     end
   end
+  context 'with custom nat_table_name' do
+    it 'no rules validate okay' do
+      pp = <<-EOS
+      class{'nftables':
+        firewalld_enable => false,
+        nat => true,
+	nat_table_name => 'mycustomtablename',
```
nbarrientos, line 120: This line could do with better indentation.

nbarrientos:
Apart from those two small comments LGTM. `REFERENCE.md` needs refreshing but this can be done later on as part of the release process.

Requesting review from @duritong

nbarrientos:
Apart from those two small comments LGTM. `REFERENCE.md` needs refreshing but this can be done later on as part of the release process.

Requesting review from @duritong

duritong (changes requested):
except for the comments from @nbarrientos I am good to go.

templates/config/puppet.nft.epp
```diff
@@ -26,6 +26,6 @@ include "custom-*.nft"
 include "inet-filter.nft"
 <% } -%>
 <% if $nat { -%>
-include "ip-nat.nft"
-include "ip6-nat.nft"
+include "ip-<%= $::nftables::nat_table_name %>.nft"
+include "ip6-<%= $::nftables::nat_table_name %>.nft"
```
duritong, line 30: Do we need to reference top scope here? Afaik this still got deprecated unless explicitly needed.

bastelfreak: yes. I didn't review the whole PR due to the size. tl;dr use `$nat_table_name` if the scope fits, otherwise `$nftables::nat_table_name`.

figless: Thanks, i've changed the variable to '$nftables::nat_table_name'

nbarrientos: LGTM. It could do with a squash.

figless: changes squashed as requested

nbarrientos (approved):

duritong (approved):

PR #115, opened 2021-12-08
Provide a mechanism to flush un-managed rules

### LINKED ISSUE #113: Correct drift between config and running rules
https://forge.puppet.com/modules/puppetlabs/firewall detects if the "running config" of iptables does not match any more with the desired configuration. It will then automatically correct this in the running config by adding or deleting the needed/unwanted rules.

Currently for puppet-nftables this is not the case. If I add or delete a rule manually with nft add/delete rule ... then puppet-nftables will not see any changes and will not correct this.

Request to have this functionality implemented.

nbarrientos:
This patchset adds a new parameter to the main class to activate a mechanism that will invoke `systemctl reload nftables` during the Puppet run if manual changes to the in-memory ruleset are detected.

To accomplish this, the systemd unit in charge of nftables is configured to write a hash of the in-memory ruleset right after starting/reloading. During the Puppet run, the hash of the current rule set is compared to the one previously stored. If the hash differs then `systemctl reload nftables` is executed to flush manual changes.

Fixes #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,13 @@ Initially puppet deploys all configuration to
 If and only if successful the configuration will be copied to
 the real locations before the service is reloaded.
 
+## Un-managed rules
+
+Rules added manually by the administrator to the in-memory ruleset
+will be left untouched. However, `nftables::allow_unmanaged_rules` can
+be set to `false` to revert this behaviour and force a reload of the
+ruleset during the Puppet run if non-managed changes are detected.
+
 ## Basic types
 
 ### nftables::config
--- a/README.md
+++ b/README.md
@@ -65,10 +65,11 @@ the real locations before the service is reloaded.
 
 ## Un-managed rules
 
-Rules added manually by the administrator to the in-memory ruleset
-will be left untouched. However, `nftables::allow_unmanaged_rules` can
-be set to `false` to revert this behaviour and force a reload of the
-ruleset during the Puppet run if non-managed changes are detected.
+By default, rules added manually by the administrator to the in-memory
+ruleset will be left untouched. However,
+`nftables::purge_unmanaged_rules` can be set to `true` to revert this
+behaviour and force a reload of the ruleset during the Puppet run if
+non-managed changes are detected.
 
 ## Basic types
 
--- a/files/systemd/nft-hash-ruleset.sh
+++ b/files/systemd/nft-hash-ruleset.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# This file is Puppet managed
+
+umask 0377
+/sbin/nft -s list ruleset | /usr/bin/sha1sum > $1
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -113,7 +113,7 @@ class nftables (
   Hash $sets = {},
   String $log_prefix = '[nftables] %<chain>s %<comment>s',
   String[1] $nat_table_name = 'nat',
-  Stdlib::Unixpath $inmem_rules_hash_file = '/var/cache/nft-memhash',
+  Stdlib::Unixpath $inmem_rules_hash_file = '/run/puppet-nft-memhash',
   Variant[Boolean[false], String] $log_limit = '3/minute burst 5 packets',
   Variant[Boolean[false], Pattern[/icmp(v6|x)? type .+|tcp reset/]] $reject_with = 'icmpx type port-unreachable',
   Variant[Boolean[false], Enum['mask']] $firewalld_enable = 'mask',
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -43,8 +43,8 @@
 # @param nat
 #   Add default tables and chains to process NAT traffic.
 #
-# @param allow_unmanaged_rules
-#   Allows to have in-memory rules that are not declared in Puppet
+# @param purge_unmanaged_rules
+#   Disallows to have in-memory rules that are not declared in Puppet
 #   code. Setting this to true activates a check that reloads nftables
 #   if the rules in memory have been modified outwith Puppet.
 #
@@ -108,7 +108,7 @@ class nftables (
   Boolean $fwd_conntrack = false,
   Boolean $inet_filter = true,
   Boolean $nat = true,
-  Boolean $allow_unmanaged_rules = true,
+  Boolean $purge_unmanaged_rules = false,
   Hash $rules = {},
   Hash $sets = {},
   String $log_prefix = '[nftables] %<chain>s %<comment>s',
@@ -175,11 +175,7 @@ class nftables (
     restart    => '/usr/bin/systemctl reload nftables',
   }
 
-  if $allow_unmanaged_rules {
-    file { $inmem_rules_hash_file:
-      ensure => absent,
-    }
-  } else {
+  if $purge_unmanaged_rules {
     exec { 'Reload nftables if there are un-managed rules':
       command     => '/usr/bin/systemctl reload nftables',
       refreshonly => false,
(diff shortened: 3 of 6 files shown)


files/systemd/nft-hash-ruleset.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
```
bastelfreak, line 2: please add a comment that this file is managed by puppet

manifests/init.pp
```diff
@@ -99,10 +108,12 @@
   Boolean $fwd_conntrack = false,
   Boolean $inet_filter = true,
   Boolean $nat = true,
+  Boolean $allow_unmanaged_rules = true,
   Hash $rules = {},
   Hash $sets = {},
   String $log_prefix = '[nftables] %<chain>s %<comment>s',
   String[1] $nat_table_name = 'nat',
+  Stdlib::Unixpath $inmem_rules_hash_file = '/var/cache/nft-memhash',
```
traylenator, line 116: Maybe  `/run/` is better location , resetting the cache on reboot sounds like a good idea and the reset can be ignored if cache is missing which will have to happen anyway.

duritong, line 116: Wouldn't something in `/run` or `/var/run` make more sense?

manifests/init.pp
```diff
@@ -164,10 +175,33 @@
     restart    => '/usr/bin/systemctl reload nftables',
   }
 
+  if $allow_unmanaged_rules {
+    file { $inmem_rules_hash_file:
```
duritong, line 179: do we really need to remove it by default?

manifests/init.pp
```diff
@@ -164,10 +175,33 @@
...
+      ensure => absent,
+    }
+  } else {
+    exec { 'Reload nftables if there are un-managed rules':
+      command     => '/usr/bin/systemctl reload nftables',
+      refreshonly => false,
+      unless      => "/usr/bin/test -s ${inmem_rules_hash_file} -a \"$(nft -s list ruleset | sha1sum)\" = \"$(cat ${inmem_rules_hash_file})\"",
```
traylenator, line 186: The `nft -s list ruleset` will also contain any chains that are present in the `nftables::noflush_chains` parameter.

Taking the typical use case of fail2ban maintaining a chain we expect that chain to drift almost constantly so ideally we want to ignore those chains when making the comparison to avoid unnecessary reloads.

The reloads themselves will be fine since these noflush chains are already ignored by the reload.

README.md
```diff
@@ -63,6 +63,13 @@ Initially puppet deploys all configuration to
 If and only if successful the configuration will be copied to
 the real locations before the service is reloaded.
 
+## Un-managed rules
+
+Rules added manually by the administrator to the in-memory ruleset
+will be left untouched. However, `nftables::allow_unmanaged_rules` can
+be set to `false` to revert this behaviour and force a reload of the
+ruleset during the Puppet run if non-managed changes are detected.
```
duritong, line 71: Should we add that this doesn't work for rulesets with counters (or does it?) and likely also not for dynamic rulesets via tools like fail2ban.

manifests/init.pp
```diff
@@ -99,10 +108,12 @@
   Boolean $fwd_conntrack = false,
   Boolean $inet_filter = true,
   Boolean $nat = true,
+  Boolean $allow_unmanaged_rules = true,
```
traylenator, line 111: `Boolean $purge_unmanaged_rules = false` is clearer to me as variable name as to what it will do. `allow_unmanaged_rules` sounds like blocking  rules from  even being added in the first place.

manifests/init.pp
nbarrientos: Maybe, I was not sure about this :)

manifests/init.pp
nbarrientos: It's mainly to remove garbage if the user switches the setting.

README.md
nbarrientos: Yup, it does work with counters as the hash is generated from the "stateless" representation of the rules. However we indeed have to decide what to do if `noflush_tables` is set.

manifests/init.pp
nbarrientos: Yup, it's probably better. I'll change it.

manifests/init.pp
nbarrientos: A hard one. The easy exit is to make them mutually exclusive, otherwise we'd have to find a way to remove the dynamic rules when Puppet takes a snapshot of what's in memory. However, having an option that purges unmanaged rules allowing unmanaged rules sounds odd to me so I'm in favour of the mutual exclusivity.

README.md
nbarrientos: Let's continue in https://github.com/voxpupuli/puppet-nftables/pull/115#discussion_r764968307, it's the same conversation.

manifests/init.pp
nbarrientos: It defaults now to `/run/puppet-nft-memhash`.

nbarrientos: It defaults now to `/run/puppet-nft-memhash`.

manifests/init.pp
nbarrientos: It's done.

manifests/init.pp
```diff
@@ -99,10 +108,12 @@
   Boolean $fwd_conntrack = false,
   Boolean $inet_filter = true,
   Boolean $nat = true,
+  Boolean $purge_unmanaged_rules = false,
   Hash $rules = {},
   Hash $sets = {},
   String $log_prefix = '[nftables] %<chain>s %<comment>s',
   String[1] $nat_table_name = 'nat',
+  Stdlib::Unixpath $inmem_rules_hash_file = '/run/puppet-nft-memhash',
```
traylenator, line 116: Apologies not to think of earlier. Should we create a directory via a systemd::tmpfile to stick this file in.

Also question if it needs to be a parameter? Why would you want to change it?

manifests/init.pp
traylenator: If we added --json to the dump filtering out the noflush chains to the dump should be doable?

traylenator: I would certainly use both options if they were not mutually exclusive.

nbarrientos: > If we added --json to the dump filtering out the noflush chains to the dump should be doable?

It's doable, however this would introduce a dependency on a JSON processor, like `jq`. Unfortunately the objects are not nested when dumped with `--json`, so the script would have to iterate over `nftables::noflush_tables` and for each element attach a filter like:

```
| del(.[] | select(.rule.table == "table_name" or .chain.table == "table_name" or .table.name == "table_name"))
```
The filter would probably have to filter by table name **and** family, as it's legal to have two tables with the same name and different families, making it even uglier.

Ugly but it should work. The script to hash the rules could be used for both dumping (to be called by Systemd as currently proposed) and comparing (to be called by the `Exec`).

Another option is to do something similar to [puppet.nft.epp](https://github.com/voxpupuli/puppet-nftables/blob/master/templates/config/puppet.nft.epp#L7) and generate an array of managed tables by removing `nftables::noflush_tables` from `$facts['nftables']['tables']`. Then sort the array, hash each table independently and then join the hashes and use that string to detect changes. Something like (pseudobash):

```bash
tables=("inet filter" "ip nat" "ip6 nat") # To be generated by Puppet
hashes=""

for table in "${tables[@]}"
do
  sha1=$(/sbin/nft -s list table $table | /usr/bin/sha1sum | cut -f 1 -d " ")
  hashes+=$sha1
done

if [ $2 == "store" ];
then
  /bin/echo $hashes > $1
elif [ $2 == "compare" ];
then
  /usr/bin/test -s $1 -a "$(cat $1)" = "$hashes"
else
  echo "Dunno what to do"
  exit 1
fi
```
We'd have to think here if the boostrapping makes sense as `$facts['nftables']['tables']` won't be available during the first run.

manifests/init.pp
nbarrientos: > Apologies not to think of earlier. Should we create a directory via a systemd::tmpfile to stick this file in.

No worries. Sounds good to me, no objection. We're already depending on the systemd module so all good. `/run/puppet-nft`?

> Also question if it needs to be a parameter? Why would you want to change it?

Thought it was a good idea to offer it but initially it could be hardcoded. If somebody ever needs to make the path customisable then merge requests are always welcome :)

JohnHolmesII:
Can this be considered again?

bastelfreak:
@JohnHolmesII if you are interested in this PR you can pick it up and rebase it.

traylenator:
Yeah the OA has other things on his mind these days.

bastelfreak:
Hi people. I think this is now implemented in https://github.com/voxpupuli/puppet-nftables/pull/253 and I'm going to close this PR.

PR #117, opened 2021-12-15
modulesync 5.1.0

bastelfreak:
modulesync 5.1.0


traylenator:
Fixed in https://github.com/voxpupuli/puppet-nftables/pull/118

PR #128, opened 2022-03-08
rspec mock systemd process on docker

traylenator:
On docker rspec the service provider is autodetected as redhat
rather than systemd.

https://tickets.puppetlabs.com/browse/PUP-11167

That causes

```
error during compilation: Parameter enable failed on Service[firewalld]: Provider redhat must have features 'maskable' to set 'enable' to 'mask' (file: /builds/ai/it
pet-module-nftables/code/spec/fixtures/modules/nftables/manifests/init.pp, line: 186)
```

mock the systemd process so that the systemd provider is used.

Also migrate existing mocks away from deprecated mocha.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -4,7 +4,6 @@
 # https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
 RSpec.configure do |c|
-  c.mock_with :mocha
 end
 
 # puppetlabs_spec_helper will set up coverage if the env variable is set.
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -4,6 +4,16 @@
 # https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
 RSpec.configure do |c|
+  c.before do
+    # select the systemd service provider even when on docker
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -4,7 +4,16 @@
 # https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
 RSpec.configure do |c|
-  c.mock_with :mocha
+  c.before do
+    # select the systemd service provider even when on docker
+    # https://tickets.puppetlabs.com/browse/PUP-11167
+    if defined?(facts) and  %w[Archlinux RedHat].include?(facts[:os]['family'])
```
bastelfreak, line 10: why limit to to those two OSes? shouldn't we do it on all?

traylenator: Was thinking someone might add slackware at some future date and then you might get the wrong service provider.
Archlinux, RedHat is everything in metadata.yml today so for sure not against changing.

nbarrientos (approved):
LGTM. Won't merge, though. Waiting for @bastelfreak's pending comment.

bastelfreak (approved):

PR #132, opened 2022-07-05
make path to `nft` binary configureable

bastelfreak:

### DIFF
--- a/data/os/RedHat.yaml
+++ b/data/os/RedHat.yaml
@@ -1,0 +1,2 @@
+---
+nftables::nft_path: /usr/sbin/nft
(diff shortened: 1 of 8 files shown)


data/os/RedHat.yaml
```diff
@@ -0,0 +1,2 @@
+---
+nftables::nft_path: /sbin/nft
```
traylenator, line 2:
```suggestion
nftables::nft_path: /usr/sbin/nft
```
since
```
rpm -ql nftables | grep bin
/usr/sbin/nft
```

bastelfreak: ugh damn :D 
I copied that path from the old code. I guess it only works bebause of the /sbin -> /usr/sbin symlink.

sebastianrakel (approved):

PR #134, opened 2022-07-05
Add Debian support

bastelfreak:
This is a rebase of https://github.com/voxpupuli/puppet-nftables/pull/66


manifests/rules/nfs3.pp
```diff
@@ -2,6 +2,7 @@
 class nftables::rules::nfs3 {
   nftables::rule {
     'default_in-nfs3':
-      content => 'meta l4proto { tcp, udp } th dport nfs accept comment "Accept NFS3"',
+      # To support nftables >= 0.9.0, but < 0.9.3, `@th,16,16 2049` is used instead of `th dport nfs`
+      content => 'meta l4proto { tcp, udp } @th,16,16 2049 accept comment "Accept NFS3"',
```
traylenator, line 6: Can we put this as parameter to the class and then set especially for the particular Debian versions.

Am always pretty nervous about changing people's rules.

Steve.

manifests/rules/out/nfs3.pp
```diff
@@ -2,6 +2,6 @@
 class nftables::rules::out::nfs3 {
   nftables::rule {
     'default_out-nfs3':
-      content => 'meta l4proto { tcp, udp } th dport nfs accept comment "Accept NFS3"',
+      content => 'meta l4proto { tcp, udp } @th,16,16 2049 accept comment "Accept NFS3"',
```
traylenator, line 5: Ditto.

traylenator (changes requested):

bastelfreak:
@traylenator this should be good to go now.

traylenator (approved):

traylenator:
Indeed much simpler sticking to newer debian.

PR #135, opened 2022-07-05
nftables service: dont explicitly set reload cmd

bastelfreak:
Puppet by default will do `systemctl restart -- nftables`. I don't see a
requirement for setting the command explicitly. Neat sideeffect is that
the path to systemctl is of course different on Debian. Dropping this
reduces complexity.


manifests/init.pp
```diff
@@ -174,10 +174,8 @@
       force   => true,
       recurse => true;
   } ~> service { 'nftables':
-    ensure     => running,
-    enable     => true,
-    hasrestart => true,
-    restart    => '/usr/bin/systemctl reload nftables',
```
traylenator, line 180: Switching to 

```
systemctl restart -- nftables
```
That is 'stop' and 'start'  which

```
ExecStop=/sbin/nft flush ruleset
ExecStart=/sbin/nft -f /etc/sysconfig/nftables.conf
```
which would definitely be bad.

traylenator (changes requested):
switching to restart which is stop and then start would definitely be bad.

manifests/init.pp
bastelfreak: oh right, that would flush the rules.

PR #138, opened 2022-07-07
Add class for pxp-agent firewalling

bastelfreak:

### DIFF
--- a/manifests/rules/pxp_agent.pp
+++ b/manifests/rules/pxp_agent.pp
@@ -1,7 +1,7 @@
 # @summary manage in pxp-agent
 # @param ports pxp server ports
 class nftables::rules::pxp_agent (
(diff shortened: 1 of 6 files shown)


manifests/rules/pxp_agent.pp
```diff
@@ -0,0 +1,10 @@
+# @summary manage in pxp-agent
+# @param ports pxp server ports
+class nftables::rules::pxp_agent (
+  Array[Stdlib::Port] $ports = [8142],
```
traylenator, line 4:
```suggestion
  Array[Stdlib::Port,1] $ports = [8142],
```

traylenator (approved):
Just the extra array length.

And thanks for the  `Array($string, true)` trick added that to https://github.com/voxpupuli/puppet-systemd/pull/288

PR #139, opened 2022-07-08
chrony: Allow filtering for outgoing NTP servers

bastelfreak:


manifests/rules/out/chrony.pp
```diff
@@ -1,7 +1,28 @@
...
-class nftables::rules::out::chrony {
-  nftables::rule {
-    'default_out-chrony':
-      content => 'udp dport 123 accept',
+# @summary manage out chrony
+# @param servers single IP-Address or array of IP-addresses from NTP servers
+class nftables::rules::out::chrony (
+  Variant[Stdlib::IP::Address,Array[Stdlib::IP::Address]] $servers = [],
+) {
+  $_servers = [$servers].flatten
```
traylenator, line 6:
```suggestion
  $_servers = Array($servers, true)
```

manifests/rules/out/chrony.pp
```diff
@@ -1,7 +1,28 @@
...
+  if empty($_servers) {
+    nftables::rule {
+      'default_out-chrony':
+        content => 'udp dport 123 accept',
+    }
+  } else {
+    $_servers.each |$index,$ps| {
+      nftables::rule {
```
traylenator, line 14: Rather than creating a rule per server you could consider creating an ip set instead?

manifests/rules/out/chrony.pp
```diff
@@ -1,7 +1,28 @@
-# manage out chrony
-class nftables::rules::out::chrony {
-  nftables::rule {
-    'default_out-chrony':
-      content => 'udp dport 123 accept',
+# @summary manage out chrony
+# @param servers single IP-Address or array of IP-addresses from NTP servers
+class nftables::rules::out::chrony (
+  Variant[Stdlib::IP::Address,Array[Stdlib::IP::Address]] $servers = [],
```
traylenator, line 4:
```suggestion
  Optional[Variant[Stdlib::IP::Address,Array[Stdlib::IP::Address]]] $servers = undef,
```
Feel it was you who encouraged me down this path generally.

Would need changes below of course.

manifests/rules/out/chrony.pp
traylenator: Not insisting since you are not going to have lots of time servers typically.

manifests/rules/out/chrony.pp
```diff
@@ -1,7 +1,28 @@
...
+    nftables::rule {
+      'default_out-chrony':
+        content => 'udp dport 123 accept',
+    }
+  } else {
+    $_servers.each |$index,$ps| {
+      nftables::rule {
+        "default_out-chrony-${index}":
+      }
+      if $ps =~ Stdlib::IP::Address::V6 {
```
traylenator, line 17: This does not follow https://puppet.com/docs/puppet/7/style_guide.html#style_guide_conditionals-simple-resource-declarations I would say.

traylenator (changes requested):

traylenator:
> @traylenator could you please take a look again? I was refactoring this while you reviewed. I now create a single rule for all IPv6 addresses and one for all IPv4 addresses. I don't think an IP set is appropriate because usually people won't have hundreds of NTP servers to talk to on the one side, and on the other the set would increase the complexity.

For sure the single rule is good.

manifests/rules/out/chrony.pp
smortex: lol, I was just thinking about making it simpler:

```suggestion
  Array[Stdlib::IP::Address] $servers = [],
```

No strong opinion about one way or another, I trend to use the simplest interface to have the less code and the less bugs and maintainability issues, but it imply more constraints for the end user of the module.

bastelfreak: I think the pure array implementation is cleaner. The idea with `Variant..` is nice if the param was initially a string but we need to support Arrays without a breaking change. Since this parameter is new, I think marking it as array is fine.

manifests/rules/out/chrony.pp
bastelfreak: can you recheck this? I think it's fine now, since all the nftable::rule resources have different titles.

spec/classes/rules/out/chrony_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      let(:facts) { os_facts }
+
+      context 'default options' do
+        it { is_expected.to compile.with_all_deps }
+        it { is_expected.to contain_nftables__rule('default_out-chrony').with_content('udp dport 123 accept') }
+        it { is_expected.to contain_concat__fragment('nftables-inet-filter-chain-default_out-rule-chrony') }
+        it { is_expected.to contain_concat__fragment('nftables-inet-filter-chain-default_out-rule-chrony_header') }
+      end
+
+      context 'with IPv4 server as string' do
```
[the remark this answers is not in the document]
bastelfreak: cleaned it up

traylenator (approved):

PR #142, opened 2022-07-26
Add all nftables families as a valid noflush pattern

luisfdez:

Nftables has a bridge family ([Bridge filtering](https://wiki.nftables.org/wiki-nftables/index.php/Bridge_filtering)) that replaces ebtables functionality.

In some case, it might be useful to avoid flushing rules defined in the associated table. However, this is not possible at the moment as the regular expression doesn't match the name.

This patch extends the regular expression defining the `noflush_tables` parameter type to accept all possible families supported by nftables as per: https://wiki.nftables.org/wiki-nftables/index.php/Nftables_families

N/A

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -116,7 +116,7 @@ class nftables (
   Variant[Boolean[false], String] $log_limit = '3/minute burst 5 packets',
   Variant[Boolean[false], Pattern[/icmp(v6|x)? type .+|tcp reset/]] $reject_with = 'icmpx type port-unreachable',
   Variant[Boolean[false], Enum['mask']] $firewalld_enable = 'mask',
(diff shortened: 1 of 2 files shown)


smortex:
A unit test that can detect regressions in future releases would be awesome!

luisfdez:
> A unit test that can detect regressions in future releases would be awesome!

Sure @smortex , will do. 

Thanks for taking a look.

luisfdez:
@smortex @traylenator 

I have added some tests and expanded the area of the patch:
* It now adds support to all the families that nftables support. Sooner of later someone will face the same situation for a missing family, so I added all of them.
* I updated the tests to iterate through all the valid types and check them. Additionally, there is a test to make sure that wrong names are not allowed.

Let me know if you have any feedback or improvements.

traylenator (approved):

PR #144, opened 2022-08-24
Properly escape bridge in rulename

### LINKED ISSUE #143: nftables::bridges creates invalid rule names when bridge devices have multiple IP addresses
<!--
Thank you for contributing to this project!

- This project has a Contributor Code of Conduct: https://voxpupuli.org/coc/
- Please check that here is no existing issue or PR that addresses your problem.
- Please fill the following form to enable us to help you.
- Our vulnerabilities reporting process is at https://voxpupuli.org/security/

-->

- Puppet: 5.5.22
- Ruby: 2.7.4p191
- Distribution: Debian Bullseye
- Module version: 2.2.1 (with a small patch to work with puppet 5.5)
- Facter: 3.14.12

Add a bridge interface with two ip addresses, eg. 
```
ip link add dev br1 type bridge
ip link set dev eth1 master br1
ip a add 172.16.1.11/24 dev br1
ip a add 172.16.1.10/32 dev br1
```

For the puppet code: 
```
# @summary Configures nftables                                                                                                                                                                                       
#                                                                                                                                                                                                                    
class profile::nftables (
  Hash[String, Hash] $rules = {},
) {
  include nftables

  $rules.each | $n, $d | {
    nftables::rule { $n:
      * => $d,
    }
  }
  Nftables::Rule <| |>
}
```

## What are you seeing

Puppet runs fail with the following error:
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Nftables::Rule[default_fwd-bridge_br1:0_br1:0]: parameter 'rulename' expects a match for Nftables::RuleName = Pattern[/^[a-zA-Z0-9_]+-[a-zA-Z0-9_]+(-\d+)?$/], got 'default_fwd-bridge_br1:0_br1:0' (file: /etc/puppet/code/environments/production/modules/nftables/manifests/bridges.pp, line: 14) on node x.test
```
The root cause of this appears to be `$facts['networking']['interfaces']` having multiple entries for the bridge device: `br1`, and `br1:0`. I understand this to be related to the multiple assigned IP addresses.

I expect the puppet run to succeed, and based what I think the intention of `nftables::bridges` is that default_fwd rules be added for the bridge device `br1`, but not `br1:0`. Please see attached output of `facter networking.interfaces`.

## Any additional information you'd like to impart

[facter-networking.interfaces.txt](https://github.com/voxpupuli/puppet-nftables/files/9402693/facter-networking.interfaces.txt)

duritong:
Fixes #143

### DIFF
--- a/manifests/bridges.pp
+++ b/manifests/bridges.pp
@@ -10,7 +10,7 @@ class nftables::bridges (
     $bridges = $interfaces.filter |$items| { $items =~ $bridgenames }
 
     $bridges.each |String $bridge| {
-      $bridge_rulename = regsubst($bridge, '-', '_', 'G')
+      $bridge_rulename = regsubst($bridge, '-|:', '_', 'G')
       nftables::rule { "default_fwd-bridge_${bridge_rulename}_${bridge_rulename}":
         order   => '08',
         content => "iifname ${bridge} oifname ${bridge} accept",
--- a/spec/classes/bridges_spec.rb
+++ b/spec/classes/bridges_spec.rb
@@ -16,6 +16,7 @@ describe 'nftables' do
                   'lo' => {},
                   'br0' => {},
                   'br1-2' => {},
+                  'br1-2:1' => {},
                 } }
           )
         end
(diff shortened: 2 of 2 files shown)


manifests/bridges.pp
```diff
@@ -10,7 +10,7 @@
     $bridges = $interfaces.filter |$items| { $items =~ $bridgenames }
 
     $bridges.each |String $bridge| {
-      $bridge_rulename = regsubst($bridge, '-', '_', 'G')
+      $bridge_rulename = regsubst($bridge, '-|:', '_', 'G')
```
smortex, line 13: For single char, I trend to prefer `[]` to `|` because it feels more readable:
```suggestion
      $bridge_rulename = regsubst($bridge, '[-:]', '_', 'G')
```
That being said, a `-` is tricky in `[]` and we have two chars only so not a big deal.

smortex (approved):
LGTM

PR #148, opened 2022-09-03
split conntrack management into dedicated classes 

duritong:
so they get consumable more easier

### DIFF
--- a/manifests/inet_filter/fwd_conntrack.pp
+++ b/manifests/inet_filter/fwd_conntrack.pp
@@ -1,0 +1,11 @@
+# @summary enable conntrack for fwd
+class nftables::inet_filter::fwd_conntrack {
+  nftables::rule {
+    'FORWARD-accept_established_related':
+      order   => '05',
+      content => 'ct state established,related accept';
+    'FORWARD-drop_invalid':
+      order   => '06',
+      content => 'ct state invalid drop';
+  }
+}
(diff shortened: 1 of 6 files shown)


manifests/inet_filter/fwd_conntrack.pp
```diff
@@ -0,0 +1,11 @@
+# enable conntrack for fwd
```
bastelfreak, line 1: this will make puppet-strings happy:

```suggestion
# @summary enable conntrack for fwd
```

spec/classes/inet_filter/fwd_conntrack_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'nftables::inet_filter::fwd_conntrack' do
+  let(:pre_condition) { 'Exec{path => "/bin"}' }
```
bastelfreak, line 6: is the pre_condition required?

manifests/inet_filter/fwd_conntrack.pp
duritong: thank you

spec/classes/inet_filter/fwd_conntrack_spec.rb
duritong: doesn't look like.

bastelfreak (approved):

PR #149, opened 2022-09-07
Allow netdev as table family in defined type nftables::chain

hugendudel:
#### Allow netdev as table family in defined type nftables::chain
Hi all, we would like to use this module in combination with the netdev table family type. 

Are there any objections?

n/a.

### DIFF
--- a/manifests/chain.pp
+++ b/manifests/chain.pp
@@ -1,7 +1,7 @@
 # manage a chain
 define nftables::chain (
   # lint:ignore:parameter_documentation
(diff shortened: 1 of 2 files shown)


traylenator:
https://wiki.nftables.org/wiki-nftables/index.php/Nftables_families#netdev

nbarrientos:
I've added an acceptance test covering the new feature.

nbarrientos (approved):
LGTM

bastelfreak (approved):

PR #153, opened 2022-10-24
Add class for outgoing HKP firewalling

bastelfreak:

### DIFF
--- a/manifests/rules/out/hkp.pp
+++ b/manifests/rules/out/hkp.pp
@@ -1,0 +1,6 @@
+# @summary allow outgoing hkp connections to gpg keyservers
+class nftables::rules::out::hkp {
+  nftables::rule { 'default_out-hkp':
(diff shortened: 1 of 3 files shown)


manifests/rules/out/hkp.pp
```diff
@@ -0,0 +1,6 @@
+# @summary allow outgoing hkp connections to gpg keyservers
+class nftables::rules::out::hkp {
+  nftables::rule { 'default_out-hkp':
+    content => 'tcp dport 11371 accept',
```
ekohl, line 4: I don't know this module very well, but how does this limit it to outgoing traffic? doesn't this also open it incoming?

bastelfreak: It parses the chain from t.he resource title. In this case the rule is added to the output chain

nbarrientos (approved):

Issue #158, opened 2023-01-26
failing to setup a basic firewall


anarcat:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 7.21.0-2
- Ruby: 3.1
- Distribution: Debian bookworm/testing
- Module version: 2.6.0

## How to reproduce (e.g Puppet code you use)

```
node default {
  include nftables::rules::out::all
  include nftables::rules::ssh
}
```

## What are you seeing

no table or ruleset created.

## What behaviour did you expect instead

some magic rules allowing outgoing connexions and incoming on port 22.

## Output log

```
anarcat@curie:nftables$ sudo pat
Info: Using environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
[...]
Info: Caching catalog for curie.anarc.at
Info: Applying configuration version '1674770130'
Warning: /Stage[main]/Nftables::Rules::Out::All/Nftables::Rule[default_out-all]/Concat::Fragment[nftables-inet-filter-chain-default_out-rule-all_header]/Concat_fragment[nftables-inet-filter-chain-default_out-rule-all_header]: Target Concat_file with path or title 'nftables-inet-filter-chain-default_out' or tag 'nftables-inet-filter-chain-default_out not found in the catalog
Warning: /Stage[main]/Nftables::Rules::Out::All/Nftables::Rule[default_out-all]/Concat::Fragment[nftables-inet-filter-chain-default_out-rule-all]/Concat_fragment[nftables-inet-filter-chain-default_out-rule-all]: Target Concat_file with path or title 'nftables-inet-filter-chain-default_out' or tag 'nftables-inet-filter-chain-default_out not found in the catalog
Warning: /Stage[main]/Nftables::Rules::Ssh/Nftables::Rule[default_in-ssh]/Concat::Fragment[nftables-inet-filter-chain-default_in-rule-ssh_header]/Concat_fragment[nftables-inet-filter-chain-default_in-rule-ssh_header]: Target Concat_file with path or title 'nftables-inet-filter-chain-default_in' or tag 'nftables-inet-filter-chain-default_in not found in the catalog
[...]

PR #162, opened 2023-01-30
README improvements

anarcat:
I couldn't figure out how to use this module when I looked at the
README. It was quickly going into pretty arcane stuff like "inet
filter" and "ip nat table" which might make sense for the module
authors or people used to nftables/iptables, but are pretty
implementation specific when coming from another networking
background.

Instead, we just explain more clearly what the module does, and
how. We also provide *more* examples, including some that might seem
obvious ("you need to include nftables first") but were not obvious to
me at all.

I also add a warning about firewalld being stopped which seems
important as well.

Closes: #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,16 +7,27 @@
 
 This module manages an opinionated nftables configuration.
 
-By default it sets up a firewall that drops every incoming
-and outgoing connection.
+By default it sets up a firewall that drops every connection, except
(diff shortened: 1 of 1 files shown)


smortex (approved):
As someone who has not seriously looked into nftables yet, this makes sense :100:

bastelfreak (approved):

duritong (approved):

nbarrientos:
Thanks for the patch!

PR #165, opened 2023-03-23
Add bridge as a valid family for chain tables

luisfdez:

Add `bridge` as a valid family to the regular expression validating the table parameter in the chain define.

Without this patch one cannot define chains belonging to tables of the `bridge` family as the validation of the `table` parameter blocks the input of names like `bridge-filter`.


bastelfreak (approved):

luisfdez:
Thanks @bastelfreak  for fixing the reference.md. It was driving me crazy.

I tried with `bundle exec rake strings:generate:reference`. 

Did you just use the `puppet strings generate` for the project?

bastelfreak:
`bundle exec rake strings:generate:reference` is fine, but you need puppet-strings 3. If you do `bundle update` it should be fine.

PR #176, opened 2023-06-03
Increased puppet/systemd upper limit to < 6.0.0

canihavethisone:
Increased puppet/systemd upper limit to < 6.0.0

Note that Metadata still shows `"version": "3.0.1-rc0"`, so the current master branch does not reflect the latest forge release

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -20,7 +20,7 @@
     },
     {
       "name": "puppet/systemd",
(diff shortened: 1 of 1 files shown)


kenyon:
The current Forge release is the same as the current master branch except for the commit setting the version number.

PR #177, opened 2023-06-19
add ldap and active directory rules

SimonHoenscheid:
Added the following rulesets:

- Ingress LDAP
- Egress LDAP
- Egress Active Directory


SimonHoenscheid:
@bastelfreak is there a chance to fix the arch Linux test?

PR #183, opened 2023-07-17
Example how to redirect one port to another

traylenator:

Add example how to redirect traffic from one port to another.

### DIFF
--- a/manifests/rule.pp
+++ b/manifests/rule.pp
@@ -13,6 +13,16 @@
 #      table   => 'ip6-nat'
 #  }
 #
+# @example Redirect port 443 to port 8443
+#  nftables::rule { 'PREROUTING-redirect':
+#    content => 'tcp dport 443 redirect to :8443',
+#    table   => 'ip-nat',
+#  }
+#  nftables::rule{'PREROUTING6-redirect':
+#    content => 'tcp dport 443 redirect to :8443',
+#    table   => 'ip6-nat',
+#  }
+#
 # @param ensure
 #   Should the rule be created.
 #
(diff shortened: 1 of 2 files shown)


manifests/rule.pp
```diff
@@ -13,6 +13,16 @@
 #      table   => 'ip6-nat'
 #  }
 #
+# @example Redirect port 443 to port 8443
+#  nftables::rule{'PREROUTING-redirect':
```
smortex, line 17: Style is not consistent with the previous examples, but none match what is recommended by [the puppet style guide](https://www.puppet.com/docs/puppet/8/style_guide.html#style_guide_module_design-spacing). Should be:

```suggestion
#  nftables::rule { 'PREROUTING-redirect':
```

smortex (commented):
That is probably fine (maybe fix style), but instead of adding examples in the README maybe it worth adding examples in the `examples` directory? This allows to have acceptance tests where we exercise the examples code to ensure it applies correctly as a catalog, ensuring that these examples are still valid if the API happen to be changed:

https://github.com/voxpupuli/voxpupuli-acceptance#examples

traylenator:
> That is probably fine (maybe fix style), but instead of adding examples in the README maybe it worth adding examples in the `examples` directory? This allows to have acceptance tests where we exercise the examples code to ensure it applies correctly as a catalog, ensuring that these examples are still valid if the API happen to be changed:
> 
> https://github.com/voxpupuli/voxpupuli-acceptance#examples

Re the examples I'll have a look next time ... else leave this here for a month or so and I'll have a go.

bastelfreak (approved):

bastelfreak (approved):

PR #189, opened 2023-08-08
nftables::simplerule::dport - takes port ranges as part of the array

### LINKED ISSUE #188: nftables::simplerule - dport will take either an array of ports *or* a port range, but not both

- Puppet: 7.19.0
- Distribution: Alma 8+9
- Module version: 3.1.0

```
class p_firewall::test () {
  $dport1 = [ 1, 2 ]
  nftables::simplerule { 'test1':
    dport => $dport1,
    proto => 'tcp',
  }

  $dport2 = '3-4'
  nftables::simplerule { 'test2':
    dport => $dport2,
    proto => 'tcp',
  }

  $dport3 = [ 1, '3-4' ]
  nftables::simplerule { 'test3':
    dport => $dport3,
    proto => 'tcp',
  }
}
```

## What are you seeing

Tests 1-2 work, test 3 fails.

nftables itself supports `tcp dport {1, 3-4} accept`, so it should work?

## Output log

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Nftables::Simplerule[test3]:
  parameter 'dport' variant 0 index 1 expects a Stdlib::Port = Integer[0, 65535] value, got String
  parameter 'dport' variant 1 expects a Stdlib::Port = Integer[0, 65535] value, got Tuple
  parameter 'dport' variant 2 expects a match for Nftables::Port::Range = Pattern[/^\d+-\d+$/], got Tuple (file: /etc/puppetlabs/.../profiles/p_firewall/manifests/test.pp, line: 15) on node XXXXXX.fnal.gov
```

## Any additional information you'd like to impart

My guess is that this just requires a change to Nftables::Port's valid data types, but my initial tests aren't sure on that.

tskirvin:

Allows nftables::simplerule's dport to accept arrays of both ports and port ranges, and not just one or the other.

Fixes #188

### DIFF
--- a/types/port.pp
+++ b/types/port.pp
@@ -1,7 +1,7 @@
 # @summary
 #   Represents a port expression to be used within a rule.
 type Nftables::Port = Variant[
-  Array[Stdlib::Port, 1],
+  Array[Variant[Nftables::Port::Range, Stdlib::Port], 1],
   Stdlib::Port,
   Nftables::Port::Range,
 ]


smortex:
> REFERENCE.md is outdated

Please update REFERENCE.md with

```sh-session
$ bundle exec rake strings:generate:reference
```

Add it to the PR and CI should continue.

tskirvin:
This is trickier on my RHEL7 build box than I was hoping; can you point me at a good getting-started page?

tskirvin:
I was able to hack my way through to the REFERENCE.md changes.

bastelfreak:
@tskirvin please rebase against our latest master branch to get rid of the merge commit.

tskirvin:
Likes like it worked properly this time!

tskirvin:
Any word on getting this merged into a full release and the forge?

types/port.pp
```diff
@@ -1,7 +1,7 @@
 # @summary
 #   Represents a port expression to be used within a rule.
 type Nftables::Port = Variant[
-  Array[Stdlib::Port, 1],
+  Array[Variant[Nftables::Port::Range, Stdlib::Port], 1],
```
traylenator, line 4: Would be good to add a couple of tests:

* And example in  of `nftables::rule` in https://github.com/voxpupuli/puppet-nftables/blob/master/spec/defines/simplerule_spec.rb
* The type itself in https://github.com/voxpupuli/puppet-nftables/blob/master/spec/type_aliases/nftables_port_spec.rb

tskirvin: I can write a quick check for the nftables::rule spec (the Rocky 9 checked failed for some reason?  There's no error message).  The mixed array example, though, seems a bit more complicated; I can look at it soon, but not today?

tskirvin: I wrote the second spec check, everything passed this time, can we get this merged up soon?

traylenator (approved):

tskirvin:
Thanks!  I'm getting this rolled out on my systems soon.

PR #191, opened 2023-08-09
Add rule to allow multicast DNS

bastelfreak:

### DIFF
--- a/manifests/rules/mdns.pp
+++ b/manifests/rules/mdns.pp
@@ -2,7 +2,10 @@
 # @summary allow incoming multicast DNS
 #
 class nftables::rules::mdns {
-  nftables::rule { 'default_in-mdns':
+  nftables::rule { 'default_in-mdns1':
     content => 'ip daddr 224.0.0.251 accept',
+  }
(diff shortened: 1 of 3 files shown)


manifests/rules/mdns.pp
```diff
@@ -0,0 +1,9 @@
+#
+# @summary allow incoming multicast DNS
+#
+class nftables::rules::mdns {
+  nftables::mdns {
+    'default_in-multicast':
+      content => 'ip daddr 224.0.0.251 accept',
```
kenyon, line 7: Formatting seemed odd

```suggestion
  nftables::mdns { 'default_in-multicast':
    content => 'ip daddr 224.0.0.251 accept',
```

kenyon: Then I saw that all of the rules are formatted like that 😢

kenyon (approved):
Legacy IP only though 😉

bastelfreak:
spotify doesn't know IPv6 :(

kenyon (approved):

PR #196, opened 2023-08-18
Inline a file instead of using a source

ekohl:
This file contains just a single file, so it's faster to include it in the catalog. It also means a cached catalog if the puppetserver is unavailable.

### DIFF
--- a/manifests/inet_filter.pp
+++ b/manifests/inet_filter.pp
@@ -9,8 +9,8 @@ class nftables::inet_filter inherits nftables {
 
   nftables::config {
     'inet-filter':
-      prefix => '',
-      source => 'puppet:///modules/nftables/config/puppet-inet-filter.nft';
+      prefix  => '',
+      content => "  include \"inet-filter-chain-*.nft\"\n";
   }
 
   nftables::chain {
--- a/files/config/puppet-inet-filter.nft
+++ b/files/config/puppet-inet-filter.nft
@@ -1,1 +1,0 @@
-  include "inet-filter-chain-*.nft"


manifests/inet_filter.pp
```diff
@@ -9,8 +9,8 @@
 
   nftables::config {
     'inet-filter':
-      prefix => '',
-      source => 'puppet:///modules/nftables/config/puppet-inet-filter.nft';
+      prefix  => '',
+      content => "  include \"inet-filter-chain-*.nft\"\n";
```
kenyon, line 13: From https://github.com/voxpupuli/puppet-nftables/blob/4e9b7fa3797a939986b50376e867a8039c5bf953/manifests/config.pp#L52C12-L52C12 it looks like this is the default.

```suggestion
```

duritong: good catch!

duritong:
Yeah, it looks like this became over time to the single line file it is now.

duritong (approved):

manifests/inet_filter.pp
ekohl: Except the newline at the end. Is that a problem? I just came across this while reviewing this module for potential usage and found this. Haven't even tried it yet.

kenyon: Looks like that newline doesn't matter here.

vchepkov:
I added #208 which adds additional line to the file

duritong:
Can we resolve the conflicts and then merge?

vchepkov:
This should be closed instead

PR #203, opened 2023-08-26
Add nftables rules for ws-discovery

bastelfreak:


ekohl (approved):
Looks correct when reading the spec.

Issue #83, opened 2021-03-23
Class[Nftables::Bridges]['bridgenames'] contains a Regexp value. It will be converted to the String '/^br.+/'


traylenator:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.20.0

Noticed in our puppetserver logs:

```
2021-03-23T07:28:07.219+01:00 WARN  [qtp1376093148-42] [puppetserver] Puppet Class[Nftables::Bridges]['bridgenames'] contains a Regexp value. It will be converted to the String '/^br.+/'
```

duritong:
I don't understand the issue. It should be a regexp and it also is: https://github.com/voxpupuli/puppet-nftables/blob/18b211e72a046db71acda7bf2fad8144b9e72e8b/manifests/bridges.pp#L5

Why is it complaining?

traylenator:
Agree , i can't see what might be wrong if anything.

traylenator:
https://tickets.puppetlabs.com/browse/PUP-7735 gives some insight.

duritong:
Can't reproduce it with the example:

```
# puppet apply foo.pp
Notice: Scope(Class[main]): Regexp[/^.$/]
Notice: Scope(Class[Foo]): Regexp[/^.$/]
Notice: Compiled catalog for xyz in environment production in 0.02 seconds
Notice: Applied catalog in 0.30 seconds
# puppet --version
7.5.0
```

And I also don't see it in our puppetserver logs.

traylenator:
@duritong that's with a Puppet 7 you do not see message ? If so suggest we ignore.

traylenator:
Ignoring as apparently not present in newer puppets.

PR #206, opened 2023-09-24
Workaround for Puppet classes not able to use rich data types

vchepkov:
This was previously reported in #83 and closed without resolution
I just added the class to apply on RHEL9 instances, and problem still exists in the current code

```
2023-09-24T13:51:33.970Z ERROR [qtp894367513-7039] [p.r.core] Internal Server Error: org.jruby.exceptions.RuntimeError: (PreformattedError) Evaluation Error: Class[Nftables::Bridges]['bridgenames'] contains a Regexp value. It will be converted to the String '/^br.+/'
```
```
# puppet --version
7.26.0
```

### DIFF
--- a/manifests/bridges.pp
+++ b/manifests/bridges.pp
@@ -2,12 +2,12 @@
 class nftables::bridges (
   # lint:ignore:parameter_documentation
   Enum['present','absent'] $ensure = 'present',
-  Regexp $bridgenames = /^br.+/
+  Pattern[/^\^.+/] $bridgenames = '^br.+',
   # lint:endignore
 ) {
   if $ensure == 'present' {
     $interfaces = keys($facts['networking']['interfaces'])
-    $bridges = $interfaces.filter |$items| { $items =~ $bridgenames }
+    $bridges = $interfaces.filter |$items| { $items =~ Regexp($bridgenames) }
 
     $bridges.each |String $bridge| {
       $bridge_rulename = regsubst($bridge, '-|:', '_', 'G')


manifests/bridges.pp
```diff
@@ -2,12 +2,12 @@
 class nftables::bridges (
   # lint:ignore:parameter_documentation
   Enum['present','absent'] $ensure = 'present',
-  Regexp $bridgenames = /^br.+/
```
traylenator, line 5: Is it documented somewhere that you can't use rich data types here ?

traylenator:
It infuriates me that I can't reproduce this outside of a real puppet server 

even using, the correct ruby:

```
steve@moving ~/G/puppet ((7.26.0))> ruby --version
jruby 9.3.9.0 (2.6.8) 2022-10-24 537cd1f8bc OpenJDK 64-Bit Server VM 25.382-b05 on 1.8.0_382-8u382-ga-1-b05 +jit [x86_64-linux]
steve@moving ~/G/puppet ((7.26.0))> bundle exec puppet --version
7.26.0
```

and 
```puppet
class abc (
  Regexp $foo = /^bar$/,
) {
  if $foo =~ Regexp {
    notify{'foo is regexp':}
  }
}
include abc
```
results in:

```
steve@moving ~/G/puppet ((7.26.0))> bundle exec puppet apply test.pp 
Notice: Compiled catalog for moving.lan in environment production in 0.15 seconds
Notice: foo is regexp
Notice: /Stage[main]/Abc/Notify[foo is regexp]/message: defined 'message' as 'foo is regexp'
Notice: Applied catalog in 0.14 seconds
```

manifests/bridges.pp
vchepkov: Lets ask @joshcooper tomorrow during #office-hours

joshcooper:
The problem is that rich data has been disabled for that catalog request, so puppetserver doesn't know how to serialize a `Regexp` instance using PSON. This can be seen from the stack trace because:

https://github.com/puppetlabs/puppet/blob/7.26.0/lib/puppet/pops/serialization/to_data_converter.rb#L195

calls:

https://github.com/puppetlabs/puppet/blob/7.26.0/lib/puppet/pops/serialization/to_data_converter.rb#L206

The "downgrade to PSON" can occur because of binary data in the catalog, often when managing kerberos keytab files.

We added a setting in puppet7 to prevent "automatic downgrade without you knowing", see https://www.puppet.com/docs/puppet/7/release_notes_puppet.html#resolved_issues_puppet_x-7-25-0-PUP-10928. It's opt-in and you'll want to set that on the puppetserver host: `puppet config set allow_pson_serialization false --section main`

Also note we changed the default to false in puppet8.

vchepkov:
@joshcooper , so I tried the suggestion and it didn't change anything. I suspect the issue is in puppet/catalog_diff
and could be related to another issue I am seeing there:

https://github.com/voxpupuli/puppet-catalog_diff/issues/105

I actually never applied the code to the node, I tried catalog_diff, saw compilation errors and went searching

joshcooper:
> I tried catalog_diff, saw compilation errors and went searching

catalog diff doesn't know how to request a rich data catalog, so it will generate a warning when trying to serialize a Regexp. This same behavior occurs when an old agent (pre 6 I think) attempts to request a catalog with rich data in it.

Note the same warning can occur if there is raw binary data in the catalog (not properly constructed Binary data types using `binary_file`).

I'm thinking this PR isn't needed as we really do want to understand and fix places where puppetserver is failing to serialize rich data. Either because puppet code should be calling `content => binary_file('path/to/file.bin')` or there's a bug causing raw binary data to "leak" into puppet (like https://github.com/puppetlabs/puppet/commit/312ab2501f8b7ae37fa8fb1ef4cebe1f7c9bb907)

vchepkov:
closing since problem lies elsewhere

bastelfreak:
@joshcooper what's the problem in catalog_diff? It has three options to get a catalog:
* via PuppetDB from `/pdb/query/v4/catalogs` https://github.com/voxpupuli/puppet-catalog_diff/blob/952b5d468676936036f9823499f5543413450b56/lib/puppet/catalog-diff/compilecatalog.rb#L56C37-L57
* via Puppetserver from `/puppet/v4/catalog` https://github.com/voxpupuli/puppet-catalog_diff/blob/952b5d468676936036f9823499f5543413450b56/lib/puppet/catalog-diff/compilecatalog.rb#L82C21-L94
* via Puppetserver from `/puppet/v3/catalog/` https://github.com/voxpupuli/puppet-catalog_diff/blob/952b5d468676936036f9823499f5543413450b56/lib/puppet/catalog-diff/compilecatalog.rb#L96C21-L96C40

I'm happy to update catalog_diff, I'm just not sure what's the problem is. @vchepkov which option did you use to request a catalog?

vchepkov:
@bastelfreak , you might have missed @joshcooper post in the ticket I have created,
https://github.com/voxpupuli/puppet-catalog_diff/issues/105#issuecomment-1736355145

I will post my command line in that ticket

traylenator:
After setting `puppet.conf` on the puppetserver with:

```ini
[main]
allow_pson_serialization  = false
```

I still have the

```
Puppet Class[Nftables::Bridges]['bridgenames'] contains a Regexp value. It will be converted to the String '/^br.+/'
```
was it meant to have gone?  (7.24.0)

traylenator:
> was it meant to have gone? (7.24.0)

and option came with 7.25 so ignore.

joshcooper:
Yep in https://github.com/puppetlabs/puppet/commit/fe63c903c2d62ae8fcb855427074a2de08f00107

duritong:
FTR: I just found this log message in question for a host, that had binary data in the catalog. Not to downgrade to pson was not possible. So I made sure that the binary data in question is read through `binary_file` and this fixed it and the catalog thus not anymore fall back to pson and the log message about Regexp does not show up anymore.

So if anybody every reports that message again, then they should look at their catalog that seems to contain binary data.

traylenator:
> So if anybody every reports that message again, then they should look at their catalog that seems to contain binary data.

Any hints how to do that ? In rspec for instance ?

PR #208, opened 2023-10-23
add ftp helper

vchepkov:
This adds ability to enable a connection tracker helper and provides typical ftp rules

### DIFF
--- a/files/config/puppet-inet-filter.nft
+++ b/files/config/puppet-inet-filter.nft
@@ -1,1 +1,2 @@
+  include "inet-filter-helper-*.nft"
   include "inet-filter-chain-*.nft"
--- a/manifests/helper.pp
+++ b/manifests/helper.pp
@@ -1,0 +1,51 @@
+# @summary manage a conntrack helper
+#
+# @example FTP helper
+#  nftables::helper { 'ftp-standard':
+#    content => 'type "ftp" protocol tcp;',
+#  }
+#
+# @param content
+#   Conntrack helper definition.
+# @param table
+#   The name of the table to add this helper to.
+# @param helper
+#   The symbolic name for the helper.
+define nftables::helper (
+  String $content,
+  Pattern[/^(ip|ip6|inet)-[a-zA-Z0-9_]+$/] $table = 'inet-filter',
+  Pattern[/^[a-zA-Z0-9_][A-z0-9_-]*$/] $helper = $title,
+) {
+  $concat_name = "nftables-${table}-helper-${helper}"
+
+  concat {
+    $concat_name:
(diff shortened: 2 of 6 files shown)


duritong:
except the missing docstrings and thus the failing tests, this looks good to me.

manifests/helper.pp
```diff
@@ -0,0 +1,41 @@
+# manage a helper
```
bastelfreak, line 1: please add puppet-strings style documentation for the parameters.

vchepkov: Done. I rebased on top of master and changed commit message to credit @jay7x , let me know if there is a better way

jay7x (approved):

files/config/puppet-inet-filter.nft
```diff
@@ -1 +1,2 @@
+  include "inet-filter-helper-*.nft"
```
duritong, line 1: Can't recall it from the top of my head, but I guess that is fine, even if that expands to 0 files - if no helper is defined?

vchepkov: correct, similar to other optional resources:
```
# grep include /etc/nftables/puppet.nft 
include "file-*.nft"
include "custom-*.nft"
include "inet-filter.nft"

```

kenyon:
The way to credit co-authors is to use `Co-authored-by: NAME <NAME@EXAMPLE.COM>`

See https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

vchepkov:
> The way to credit co-authors is to use `Co-authored-by: NAME <NAME@EXAMPLE.COM>`
> 
> See https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

Thank you, amended

manifests/helper.pp
```diff
@@ -0,0 +1,53 @@
...
+#  }
+#
+# @param content
+#   Conntrack helper definition.
+# @param table
+#   The name of the table to add this helper to.
+# @param helper
+#   The symbolic name for the helper.
+define nftables::helper (
+  # lint:ignore:parameter_documentation
```
ekohl, line 15: Why is this needed? Looks like you have documented them.

vchepkov: I copied it from an existing define. let me drop it

ekohl: Did you forget to push your changes?

vchepkov: it seems that way. sigh. Hopefully that time is a charm

PR #213, opened 2023-11-17
allow puppet/systemd v6

vchepkov:


vchepkov:
@bastelfreak , could we merge this before the release MR is merged, please ?

PR #217, opened 2023-12-02
Add rule for filtering outgoing DNS server traffic

bastelfreak:

### DIFF
--- a/manifests/rules/out/dnsserver.pp
+++ b/manifests/rules/out/dnsserver.pp
@@ -1,0 +1,30 @@
+#
+# @summary manage outgoing DNS responses from a DNS server
+#
+# @param dns_servers optional list of local ip addresses from the DNS server
+#
+class nftables::rules::out::dnsserver (
+  Array[Stdlib::IP::Address] $dns_servers = [],
+) {
+  unless empty($dns_servers) {
+    $dns_servers.each |$index,$dns| {
+      $content = $dns ? {
+        Stdlib::IP::Address::V6 => "ip6 saddr ${dns}",
+        Stdlib::IP::Address::V4 => "ip saddr ${dns}",
+      }
+      nftables::rule { "default_out-dnsservertcp-${index}":
+        content => "${content} tcp sport 53 accept",
+      }
+      nftables::rule { "default_out-dnsserverudp-${index}":
+        content => "${content} udp sport 53 accept",
+      }
+    }
+  } else {
+    nftables::rule {
+      'default_out-dnsserverudp':
+        content => 'udp sport 53 accept';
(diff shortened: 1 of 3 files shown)


manifests/rules/out/dnsserver.pp
```diff
@@ -0,0 +1,45 @@
...
+        }
+      } else {
+        Nftables::Rule["default_out-dnsserverudp-${index}"] {
+          content => "ip daddr ${dns} udp sport 53 accept",
+        }
+      }
+
+      nftables::rule {
+        "default_out-dnsservertcp-${index}":
+      }
```
smortex, line 26: I guess it follow a structure as seen in other classes, but that feels quite convoluted (altering a resource that appears only after in the manifest) and I would have expected something like this that feels more simple:

```puppet
      $content = $dns ? {
        Stdlib::IP::Address::V6 => "ip6 daddr ${dns} udp sport 53 accept",
        Stdlib::IP::Address::V4 => "ip daddr ${dns} udp sport 53 accept",
      }

      nftables::rule { "default_out-dnsservertcp-${index}":
        content => $content,
      }
```

I have not checked all other rules classes, maybe such a construct make sense in some cases.  But I am not sure it is better to follow that pattern when a simpler construct can be used?

smortex (approved):
I am a bit worried about the way this feels complex, but understand that consistency is sometimes better that simplicity, so maybe this is fine?  That's clearly _not_ a no-go for me, so I'll approve this, but maybe users of the module will want to share their mind.

manifests/rules/out/dnsserver.pp
bastelfreak: I am not a fan of this contruct either, but I copied it from the rules::out::dns class and wanted to stay consistent.

manifests/rules/out/dnsserver.pp
```diff
@@ -0,0 +1,30 @@
...
+  Array[Stdlib::IP::Address] $dns_servers = [],
+) {
+  unless empty($dns_servers) {
+    $dns_servers.each |$index,$dns| {
+      $content = $dns ? {
+        Stdlib::IP::Address::V6 => "ip6 saddr ${dns}",
+        Stdlib::IP::Address::V4 => "ip saddr ${dns}",
+      }
+      nftables::rule { "default_out-dnsservertcp-${index}":
+        content => "${content} tcp sport 53 accept",
```
bastelfreak, line 16: derp, I need to rethink this. DNS servers don't respond with src port 53 but a random port. Maybe we can just close the PR.

duritong:
I think this was essentially done with #222

PR #222, opened 2023-12-19
rules::out:dns: refactor for better readability

bastelfreak:
this got cherry-picked from https://github.com/voxpupuli/puppet-nftables/pull/217/


manifests/rules/out/dns.pp
```diff
@@ -1,34 +1,19 @@
...
 # @param dns_server specify dns_server name
 class nftables::rules::out::dns (
-  Optional[Variant[String,Array[String,1]]] $dns_server = undef,
+  Array[Stdlib::IP::Address] $dns_server = [],
 ) {
-  if $dns_server {
-    any2array($dns_server).each |$index,$dns| {
-      nftables::rule {
-        "default_out-dnsudp-${index}":
+  unless empty($dns_server) {
```
vchepkov, line 6: This is probably redundant, each will not create any resources with empty array, right?

bastelfreak: yes. but I need the condition because of the else block below.

vchepkov (approved):

PR #233, opened 2023-12-30
rules::out::mdns & rules::mdns: Allow interface filtering

bastelfreak:


smortex (approved):
That looks reasonable.

bastelfreak:
added unit tests

PR #234, opened 2023-12-31
rules::ospf3 & rules::out::ospf3: Allow filtering on outgoing interfaces

bastelfreak:
Also contains https://github.com/voxpupuli/puppet-nftables/pull/233


smortex (approved):
That also look reasonable.  Maybe adding tests could help if the syntax ever change?

bastelfreak:
added unit tests

PR #242, opened 2024-04-26
update puppet-systemd upper bound to 8.0.0

TheMeier:


duritong (approved):
lgtm

TheMeier:
Whats with the failing tests?

duritong:
It looks like nftables segfaults when starting, which I think is rather a test flake than something that is caused by the new module.

PR #247, opened 2024-06-26
New clobber_default_config paramater

traylenator:

Certain OSes namely Debian and Archlinux provide default rules with the OS.

This module has always respected those rules and appended all of its own rules to the end of the existing rules.

The new parameter `clobber_default_config` if set `true` (default `false`) will drop any existing OS provided rules.

Also related to acceptance tests only on Archlinux where the default OS provided configuration requires kernel >= 6.3 we purge the default rules if required.


traylenator:
Along with #245 (centos8) and #246 (debian11) this should green up the tests.

duritong (approved):

PR #250, opened 2024-07-09
modulesync 9.4.0

bastelfreak:
modulesync 9.1.0


bastelfreak:
Due to Puppet 7 still relying on legacy facts, despite the include_legacy_facts parameter that can be set to false, this fails. See https://github.com/puppetlabs/puppet/pull/9427

traylenator:
Tried generating a fresh REFERENCE.md but can't get a diff of any sort.

kenyon:
@traylenator I updated `REFERENCE.md`, but looks like there are actual test failures now.

traylenator:
> @traylenator I updated `REFERENCE.md`, but looks like there are actual test failures now.

I've not had a chance to check yet but think it may be what I mention here.

https://github.com/voxpupuli/puppet-nftables/pull/266#issuecomment-2636440785

traylenator:
Installing "netbase" on Ubuntu 20.04 adds the necessary `/etc/services` file. Should probably be one in the module itself but some what easier after this.

PR #251, opened 2024-07-28
Release prep for v3.7.2 - Generated changelog

canihavethisone:
Release prep for v3.7.2 - Generated changelog


kenyon (changes requested):
It would be version 4.0.0 since there is a breaking change. We should get #250 fixed and merged before releasing too.

canihavethisone:
closing this while I submit a PR to update the unmanaged rules purge consideration

PR #253, opened 2024-08-03
Reload nftables service if hash at last service load does not match

canihavethisone:
This is a variation on https://github.com/voxpupuli/puppet-nftables/pull/115, where I use:
- 2 execs
- no script file on disk
- no service file modifications.

Effect:

**First install:**
- after nftables is installed and configured and service started, a hash of memory state (without counters) is written to disk in the same puppet run

**Subsequent runs:**
_Hash file not present (eg deleted):_
1. nftables service is notified to reload
2. hash file is regenerated

_Hash of memory state does not match hash on disk (eg rule manually added in memory):_
1. nftables service is notified to reload
2. hash file is regenerated

_Changes invoked by Puppet to rules:_
1. hash file is regenerated (as this module already notifies the service)

**Exec time hit**
Everyone tries not to use execs, but they work well in this case and over a number of tests I barely saw a `0.1 sec` difference, and that wasn't consistent so I think its a good solution.

Thanks for your consideration of this PR!

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -235,16 +235,18 @@ class nftables (
   if $purge_unmanaged_rules {
     # Reload the nftables ruleset from the on-disk ruleset if there are differences or it is absent. -s must be used to ignore counters
     exec { 'nftables_memory_state_check':
-      command => ['echo', 'reloading_nftables'],
-      path    => $facts['path'],
-      unless  => ["test -s ${inmem_rules_hash_file} -a \"$(nft -s list ruleset | sha1sum)\" = \"$(cat ${inmem_rules_hash_file})\""],
-      notify  => Service['nftables'],
+      command  => ['echo', 'reloading_nftables'],
+      path     => $facts['path'],
+      provider => shell,
+      unless   => ["test -s ${inmem_rules_hash_file} -a \"$(nft -s list ruleset | sha1sum)\" = \"$(cat ${inmem_rules_hash_file})\""],
+      notify   => Service['nftables'],
     }
 
     # Generate nftables_hash upon any changes from the nftables service 
     exec { 'nftables_generate_hash':
-      command     => "nft -s list ruleset | sha1sum > ${inmem_rules_hash_file}",
+      command     => ["nft -s list ruleset | sha1sum > ${inmem_rules_hash_file}"],
       path        => $facts['path'],
+      provider    => shell,
       subscribe   => Service['nftables'],
       refreshonly => true,
     }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -233,7 +233,7 @@ class nftables (
   }
 
   if $purge_unmanaged_rules {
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -221,6 +232,24 @@
     restart    => 'PATH=/usr/bin:/bin systemctl reload nftables',
   }
 
+  if $purge_unmanaged_rules {
+    # Reload the nftables ruleset from the on-disk ruleset if there are differences or it is absent. -s must be used to ignore counters
+    exec { 'nftables_memory_state_check':
+      command => 'echo "reloading nftables"',
+      path    => ['/usr/sbin', '/sbin', '/usr/bin', '/bin'],
```
bastelfreak, line 239: To make this more generic, please use:
```suggestion
      path    => $facts['path'],
```

manifests/init.pp
```diff
@@ -221,6 +232,24 @@
...
+      unless  => "/usr/bin/test -s ${inmem_rules_hash_file} -a \"$(nft -s list ruleset | sha1sum)\" = \"$(cat ${inmem_rules_hash_file})\"",
```
bastelfreak, line 240: can you please update the command and unless parameter to be arrays and not strings? That's a new best practice. Perforce already has a puppet-lint plugin for it and we will probably roll that out in the future as well.

canihavethisone: this was tricky with the length of the unless.  I have put it in an array but all as 1 element, it failed with splitting it up in my testing.  The second exec to generate the has I couldn't get to work in an array at all, even as 1 element.  Any suggestions?

bastelfreak:
Please run `bundle exec rake strings:generate:reference` to update the REFERENCE.md

spec/classes/nftables_spec.rb
```diff
@@ -298,6 +310,31 @@
...
 
+      context 'when purging unmanaged rules' do
+        let(:params) do
+          {
+            'purge_unmanaged_rules' => true,
+            'inmem_rules_hash_file' => '/foo/bar',
+          }
+        end
+
+        it { is_expected.not_to contain_file('/foo/bar') }
```
bastelfreak, line 321: you can probably please rubocop by running `bundle exec rake rubocop:autocorrect`

canihavethisone: Ran autocorrect, thanks.

manifests/init.pp
canihavethisone: Thanks for the tip to add shell provider, that has allowed the second exec to be in an array.  I tried breaking the execs into elements, but couldn;t get them to work in various permutations.  I think they will need to remain as single elements.

canihavethisone:
one question I have relates to the choice of location for the hash by the previous PR, which I have copied.  Writing it to `/run` means that upon a reboot the file is missing and will be regenreated, notifying the nftables service to reload unneccesarily.  Should this default to a location where the hash file survives a reboot, such as `/var/tmp`?

PR #258, opened 2024-09-14
Support logging to NFLOG group

deric:
With specified `nftables::log_group` log messages will be send to appropriate `NFLOG` group.

The options `flags` and `group` are mutually exclusive.

with 
```yaml
nftables::log_group: 1
```
the rule should look like this:
```
limit rate 3/minute log prefix "[nftables] OUTPUT Rejected: " group 1
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -72,6 +72,10 @@
 #  to the rules that log discarded traffic. Set to false to
 #  disable rate limiting.
 #
+# @param log_group
+#  When specified, the Linux kernel will pass the packet to nfnetlink_log
+#  which will send the log through a netlink socket to the specified group.
+#
 # @param reject_with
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -147,6 +151,7 @@
   Hash $rules = {},
   Hash $sets = {},
   String $log_prefix = '[nftables] %<chain>s %<comment>s',
+  Optional[Integer] $log_group = undef,
```
traylenator, line 154:
```suggestion
  Optional[Integer[0]] $log_group = undef,
```
Presumably >= 0

deric:
@traylenator The type has been fixed. Any other issues?

traylenator:
Sorry just noticed, will look later.

traylenator (approved):

PR #263, opened 2024-10-10
file: Fix heredoc example

deric:
Provided example would fail on syntax validation
```
Error: Could not parse for environment production: Heredoc without end-tagged line
```

### DIFF
--- a/manifests/file.pp
+++ b/manifests/file.pp
@@ -1,7 +1,7 @@
 # @summary Insert a file into the nftables configuration
 # @example Include a file that includes other files
 #   nftables::file{'geoip':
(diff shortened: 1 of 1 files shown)


manifests/file.pp
```diff
@@ -1,10 +1,10 @@
 # @summary Insert a file into the nftables configuration
 # @example Include a file that includes other files
 #   nftables::file{'geoip':
-#     content => @(EOT)
+#     content => @("EOT"),
```
kenyon, line 4: Interpolation not needed in this heredoc:

```suggestion
#     content => @(EOT),
```

PR #266, opened 2024-12-14
Allow puppet-systemd 8.x

jay7x:
- **puppet/systemd: allow 8.x**

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -20,7 +20,7 @@
     },
     {
       "name": "puppet/systemd",
-      "version_requirement": ">= 2.0.0 < 8.0.0"
+      "version_requirement": ">= 2.0.0 < 9.0.0"
     },
     {
       "name": "puppetlabs/stdlib",


duritong:
Why are tests always failing?

traylenator:
> Why are tests always failing?

```
223) nftables::simplerule on rocky-9-x86_64 only IPv4 TCP traffic is expected to compile into a catalogue without dependency cycles
       Failure/Error: it { is_expected.to compile }
         error during compilation: Parameter enable failed on Service[firewalld]: Provider debian must have features 'maskable' to set 'enable' to 'mask' (file: /scratch/actions/actions-runner-2.305.0/instance14/_work/puppet-nftables/puppet-nftables/spec/fixtures/modules/nftables/manifests/init.pp, line: 267)
       # ./spec/defines/simplerule_spec.rb:150:in `block (5 levels) in <top (required)>'
 ```
 
 This is when we get the wrong service provider "debian" and not "systemd'.
 
Thought we has  some mocking somewhere to satisfy:
 ```
 confine :true => Puppet::FileSystem.exist?('/proc/1/comm') && Puppet::FileSystem.read('/proc/1/comm').include?('systemd')
 ```
 in the service systemd provider. Failing to find it.

traylenator:
> Thought we has some mocking somewhere to satisfy:

Yeah we seemed to have lost here:

https://github.com/voxpupuli/puppet-nftables/commit/65ed81ba2f490e8016946b7f4923bf9f276d808e#diff-89eebfcbc0f14b6d989517837ca1e94fce4e2ce9a03233641cd936f2b8d2ed94

kenyon (approved):

PR #272, opened 2025-02-28
Add firewall rule for incoming rsync requests

bastelfreak:


manifests/rules/rsync.pp
```diff
@@ -0,0 +1,7 @@
+# @summary manage in http
```
TheMeier, line 1: that does not look right
```suggestion
# @summary allow rsync connections
```

TheMeier (changes requested):

TheMeier (approved):

PR #288, opened 2025-10-27
Add nftables::rules::mosh

Sharpie:

This commit adds a class for managing firewall input rules for mosh, the mobile shell. Mosh bootstraps over SSH and then starts a server process that listens on a UDP port between 60000 and 61000.

See: https://mosh.org/#usage

N/A


Sharpie:
Added missing items that were causing tests to fail.

duritong (approved):

Issue #155, opened 2022-10-29
default debian config not overwritten -> noflush_tables not working


dmaes:
## How to reproduce (e.g Puppet code you use)
Use this module on Debian, optionally set the `noflush_tables` parameter

## What are you seeing
Despite setting `noflust_tables`, all my tables where flushed when doing a `systemctl reload nftables`.
Upon further investigation, I saw that, on Debian, the default config path `/etc/nftables.conf`is used in the systemd service overwrite,
and that `file_line` is used to add `include "/etc/nftables/puppet.nft"` to include all the rules from the puppet-managed files.

The problem is, Debian has the following config by default in `/etc/nftables.conf`:
```
#!/usr/sbin/nft -f

flush ruleset

table inet filter {
	chain input {
		type filter hook input priority 0;
	}
	chain forward {
		type filter hook forward priority 0;
	}
	chain output {
		type filter hook output priority 0;
	}
}
```

This does not get cleaned up, only the `include ...` line is added at the end, meaning you will _always_ flush all tables and create a default `inet filter` table and chains, before you get to the puppet-managed part. 

I think either this file needs to be cleaned up by the module, or a new, empty, file needs to be used.
Else we get unexpected results, like `noflush_tables` not working,
and possibly other things if Debian maintainers decide to add other defaults to a file that we use, but don't fully manage...

traylenator:
New parameter added to allow clobber of default rules.

PR #307, opened 2026-06-26
Add more info that the default config is still in use

TrayserCassa:
Hey, it's bin a while since I used Puppet so this can be seen from a beginner view.

I spend some time debugging this and I feel a bit embarrassed. Nevertheless I think we should add small hint in the documentation that the default os config file didn't get overwritten by puppet.

Maybe the README change is a bit to much, but for me that was the starting point for this module. 

It would save me some hours/frustration and also some tokens if this would be mentioned in the Documentation. To be fair, it is in the docs, but a small hint in the `purge_unmanaged_rules` would be really helpful I think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,20 @@ and/or `nftables::nat` to `false` and build your whole nftables
 configuration from scratch by using the building blocks provided by
 this module. Look at `nftables::inet_filter` for inspiration.
 
+> **Warning: the default OS config may flush your un-managed tables.**
+> On most distributions the package ships a default `/etc/nftables.conf`
+> that begins with `flush ruleset`. By default this module does **not**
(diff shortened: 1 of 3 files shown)


kenyon (changes requested):
`REFERENCE.md` is autogenerated from the source code, so you'll need to update the corresponding pp file and regenerate `REFERENCE.md` with `bundle exec rake strings:generate:reference`.

kenyon (approved):

traylenator:
It's too long for me to remember here but Debian and RHEL are different here.

Debian does have default rules - RHEL does not I think.

 I think there will be some  history in 

https://github.com/voxpupuli/puppet-nftables/issues/155

Adding more explanations - always good.

